PR #26, opened 2018-04-12
Manage dependencies with dep

rargulati:
This PR addresses the issue brought forth in https://github.com/uber/prototool/issues/4

Each commit represents a single action take to get to the next state.

Of note, dep upgraded the following dependencies:
    - [uber/atomic](https://github.com/uber-go/atomic) ([diff](https://github.com/uber-go/atomic/compare/54f72d32435d760d5604f17a82e2435b28dc4ba5...8474b86a5a6f79c443ce4b2992817ff32cf208b8))
    - [russross/blackfriday](https://github.com/russross/blackfriday) ([diff](https://github.com/russross/blackfriday/compare/cadec560ec52d93835bf2f15bd794700d3a2473b...55d61fa8aa702f59229e6cff85793c22e580eaf5))
    - [davecgh/go-spew](https://github.com/davecgh/go-spew) ([diff](https://github.com/davecgh/go-spew/compare/346938d642f2ec3594ed81d874461961cd0faa76...8991bc29aa16c548c550c7ff78260e27b9ab7c73))

As all tests passed (on running `make test`, I kept them as is. If requested, I can updated `Gopkg.toml` to have the commits referred to in `glide.lock`.

Thanks for releasing this! Looking forward to making use of it and hopefully helping out where I can!

Issue #4, opened 2018-04-10
Dependency Management


peter-edge:
To get this off the ground, I did what I was used to and used [Glide](https://github.com/Masterminds/glide) as my dependency management tool. I wanted this to be go-gettable, so all this does is specify all necessary dependencies without versions, which are then checked into `vendor`. This has been fine for development, as Prototool does not expose any external API, and this makes installing Prototool as easy as `go get github.com/uber/prototool/cmd/prototool`.

However, we eventually will want to expose parts of Prototool as a Golang library, and this isn't a very efficient way to do this. We probably don't want to check in `vendor` in this case, and we want to get up to date by using [Dep](https://github.com/golang/dep). Installing Prototool will instead move to curl'ing binaries released to GitHub Releases.

Update: No longer checking in vendor, so probably all that is left here is to move from Glide to Dep.

breerly:
Hey @peter-edge, I think we want to revisit moving prototool to dep.

It *looks* like Go is going to be throwing away dep in favor of vgo:

https://github.com/golang/go/issues/24301

Rather than switch to dep now, I say we hold back a bit to see what happens w/ vgo.

If Glide is that much of a pain, then we can just check in the vendor repo.

Regardless, Prototool is a binary, and so we shouldnt need this for v1.0.

cc @blampe @akshayjshah

peter-edge:
Sounds fine to me.

peter-edge:
Glide is fine, I used to check in vendor but especially pre-1.0 I don’t want to, it significantly increases the size of each commit that has a vendor update and makes making this a library harder.

breerly:
Kk, using Glide for development seems fine for now!

peter-edge:
Let’s keep this open for now to track this and to show users the rationale, I know others have a different opinion.

johanbrandhorst:
I personally think it's a mistake to forsake `dep` in favor of waiting for `vgo`. `vgo` might make it into 1.11 as an optional extension, and adoption will probably take a while. Meanwhile, `dep` has been very well received by the community, is very similar to existing dependency management solutions (npm, yarn, cargo), will provide a clear upgrade path to the `vgo` system, and works _right now_. I get the point about this being released as a binary primarily, but if people want to use a stable version of this, they may well use `dep` to track the version, and I think we could benefit from having these definitions in this repo right now.

blampe:
FWIW `vgo` already supports `dep` manifests out of the box, so I don't think we lose anything by doing this.

[...]

PR #41, opened 2018-04-23
glide: convert to dep

jzelinskie:
This maintains the default branch for all dependencies rather than the
default dep behavior which is finding stable tags.

Closes #26.


breerly:
@jzelinskie thanks for the PR!

We're going to close this PR for now because it *looks* like Golang will be adopting something called [vgo](https://github.com/golang/go/issues/24301) instead of dep!

Check out #4 for the rationale.

PR #47, opened 2018-04-30
Move internal/x/wkt to internal/wkt

peter-edge:
This is to review the `wkt` package, which is very easy. A package comment was added for more explanation.


AlexAPB:
Can we change `Filenames` to the `map[string]struct{}` and drop `init()` and `FilenameMap`? Browsing through the codebase it looks like we don't rely on `Filenames` being an ordered list so this would simplify things.

peter-edge:
Yea we can do that. Funny enough, the returned `text.Failures` from `wktDirectlyImportedChecker` will be sorted regardless because the output messages use the filename and message is the final sort field for a `text.Failure` haha.

PR #49, opened 2018-04-30
Move internal/x/cfginit to internal/cfginit

peter-edge:
This is to review the `cfginit` package, which is very easy. A package comment was added for more explanation.


peter-edge:
How about I also make two public functions, one for commented out and one
not?

peter-edge:
@AlexAPB I went to change `uncomment` to `commentOut`, and realized that in https://github.com/uber/prototool/blob/dev/internal/x/exec/runner.go#L132 we have it as `uncomment` as well, which makes this awkward. This is intentional:

```
prototool init
```

This generates the config file you usually want.

```
prototool init --uncomment
```

This generates the config file with everything uncommented, ie this is the non-default option. I'd want the internal code to match this for less confusion.

Theres a few things we could do here:

- Change the boolean to `commentOut` in `cfginit` anyway, and then have `!uncomment` in the `Runner`
- Make two public functions, `Generate` and `GenerateUncommented`, and then switch on the boolean in the `Runner`
- Change the flag name to something else? I'm not sure what.

Thoughts?

AlexAPB:
hmm, yeah I can buy keeping it the same since it reflects the flags users pass in. I'm not sure what could make it better -- maybe @amckinney has thoughts on this.

In any case it was just a nit. It's all internal so changing it may be too much of a hassle for a minor readability change.

You've got my stamp 👍

AlexAPB:
what if we renamed `cfginit` to `cfgtemplate` or even just `template`?

AlexAPB:
yeah leaving it as is is fine for me since it's internal

AlexAPB:
also you have conflicts on this branch now

peter-edge:
There’s going to be a lot of conflicts as we review these packages, I’m not going to be aggressive in updating them as I intend to leave these PRs up for a while if that sounds good.

AlexAPB:
yup sounds good!

amckinney:
I like the consistency between the command-line flag `uncomment` and the implementation detail, so I think we should leave it as is.

amckinney:
In lines 36-38:
```
# Paths to exclude when using directory mode.
# These are prefixes, not regexes, so path/to/a will ignore anything beginning with
# $(dirname some/dir/prototool.yaml)/path/to/a including for example $(dirname some/dir/prototool.yaml)/path/to/ab.
```

Ignoring directory `ab` seems like a bug. Can we fix this so that ignoring `a` won't ignore `a*`, but only `a/*`?

peter-edge:
This is really more of a question for when we review internal/x/settings, which parses the config files. I personally like that these are prefixes, not regexes, but we can change that in internal/x/settings. This package is just explaining the current behavior in comments on a newly-generated `prototool.yaml` file. Does that make sense?

amckinney:
In lines 43-44:
```
# Do not use the default exclude paths.
# The only default exclude path is "vendor".
```

Do we ever see this growing past `vendor`? It might be more clear / easier to digest if we simply have `include_vendor: true`, or something similar.

peter-edge:
We might want to remove this default exclude functionality altogether actually, but another question for internal/x/settings.

peter-edge:
Dependent on #77.

PR #50, opened 2018-04-30
Move internal/x/strs to internal/strs

peter-edge:
This is to review the `strs` package, which is very easy. A package comment was added for more explanation.


AlexAPB:
For the next set of reviews, it'd be helpful to put up PRs that have two commits. One that deletes the `/x` package and the second that places it in its final spot. Just moving the packages prevents reviewers from making in-line comments.

peter-edge:
I could even just redo these and do that if it’s that big a thing.

I think if you expand the files you can make comments regardless?

AlexAPB:
That's what I thought too, but it won't let you after you expand too many lines. Also for straight file name changes on other PRs, you can't expand the file at all :(

peter-edge:
I can redo these if it’ll help that much.

AlexAPB:
Nahh, let's just do that moving forward. These PRs look small enough so far

AlexAPB:
- `IsCamelCase` doesn't mention it deals with numbers
- `IsCamelCase` returns true for strings that start with numbers

nit since internal:
- for tests taking in a modifer, should take in multiple instead? Not sure if this would have been more helpful
eg `func DedupeSlice(s []string, modifiers ...func(string) string) []string { ... } `
- `extraRunes` may be more intuitive as `ignoreRunes`

tests
- lets replace `assert.Equal()` where we test against a bool, with its corresponding `assert.True()` or `assert.False()`
- missing tests for `IsUpperSnakeCase`, 'IsLowercase', 'IsUppercase', `ToUpperSnakeCase` and `DedupeSortSlice`
- tests don't test for empty string cases
- intersectionSlice should comment saying empty strings are dropped

peter-edge:
Why replace equal with true or false? With equal, you see the expected value, and the actual. That seems strictly better, no?

Ie “expected 2, got 3” instead of “expected true, got false”. Can you explain?

peter-edge:
That makes sense!

peter-edge:
- Updated the documentation on `IsCamelCase`.
- It is OK if something starts with a number for `IsCamelCase`, this is in the documentation now though.
- There's no practical benefit to taking in multiple `modifiers` as I never use it, and this would just be slower since you're iterating through a loop and doing more updates/potentially allocations. This is an internal package so we're not locked into the API here.
- Personally I think `extraRunes` makes more sense because its "extra runes that are ok in addition to AZaz09". But if you feel otherwise we can change it.
- Updated the tests to use `assert.True` and `assert.False`.
- Updated the documentation to note where empty strings are dropped.
- Added tests for empty strings and that the dedupe and intersection functions ignore empty strings.

Will add tests for the missing functions now. Thanks for the detailed feedback!

peter-edge:
Added tests for IsUpperSnakeCase, IsLowercase, IsUppercase, ToUpperSnakeCase and DedupeSortSlice.

amckinney:
Should we replace `isUppercaseRune` with `unicode.IsUpper`? There might be other valid unicode uppercase characters allowed in Proto besides the traditional A-Z.

peter-edge:
Done.

amckinney:
`DedupeSlice` seems to be doing more than necessary (i.e. applying the modifier > 1 & unnecessarily checking for `""`) Can we rewrite to:
```
// DedupeSlice returns s with no duplicates and no empty strings, in the same order.
// If modifier is not nil, modifier will be applied to each element in s.
func DedupeSlice(s []string, modifier func(string) string) []string {
	m := make(map[string]struct{}, len(s))
	o := make([]string, len(s))
	for _, e := range s {
		if e == "" {
			continue
		}
		key := e
		if modifier != nil {
			key = modifier(e)
		}
		if _, ok := m[key]; ok {
			continue
		}
		m[key] = struct{}{}
		o = append(o, key)
	}
	return o
}
```

PR #51, opened 2018-04-30
Move internal/x/cmd to internal/cmd

peter-edge:
This is to review the `cmd` package. This package contains the logic that sets up Prototool with github.com/spf13/cobra. `cmd/prototool` calls into this package, which calls into `exec` to execute the logic through `exec.Runner` in a cobra-independent fashion.


AlexAPB:
Since there's a lot going on here can you update this diff with delete/add commits? I know most of it are test files, but it'd still help!

AlexAPB:
- Let's alphabetize the command declarations, flag fields, flag methods, etc. so it's easier to traverse.
- Can we move `flags` and it's associated methods to a new `flags.go` file? `cmd.go` is fairly large.

peter-edge:
Closing in favor of #71.

PR #52, opened 2018-05-01
Move internal/x/diff to internal/diff

peter-edge:
This is probably the simplest package of all of them. This is just adapted from https://github.com/golang/go/blob/e86168430f0aab8f971763e4b00c2aae7bec55f0/src/cmd/gofmt/gofmt.go to get the diff functionality for `prototool format -d`. I did this in the two commit style per #50.


peter-edge:
I added extremely basic testing as well.

internal/diff/diff.go
```diff
@@ -0,0 +1,109 @@
...
+	file, err := ioutil.TempFile(dir, prefix)
+	if err != nil {
+		return "", err
+	}
+	_, err = file.Write(data)
+	if err1 := file.Close(); err == nil {
+		err = err1
+	}
+	if err != nil {
+		defer func() { _ = os.Remove(file.Name()) }()
```
AlexAPB: nit: does this need to be in a defer since we're returning immediately after?

AlexAPB:
Thanks for doing the two commit thing so we can add in-line comments!
It looks like it shows as outdated (unfortunately makes sense) so you'll have to click through them to see the comments

internal/diff/diff.go
peter-edge: Nope, deleting the defer.

AlexAPB:
👍

internal/diff/diff.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// replaceTempFilename replaces temporary filenames in diff with actual one.
+//
+// --- /tmp/gofmt316145376	2017-02-03 19:13:00.280468375 -0500
+// +++ /tmp/gofmt617882815	2017-02-03 19:13:00.280468375 -0500
+// ...
+// ->
+// --- path/to/file.go.orig	2017-02-03 19:13:00.280468375 -0500
+// +++ path/to/file.go	2017-02-03 19:13:00.280468375 -0500
```
amckinney: nit: Can we indent the `2017-02-03 19:13:00.280468375 -0500` so it lines up nicely?

internal/diff/diff.go
```diff
@@ -0,0 +1,109 @@
...
+
+	cmd := "diff"
+	if runtime.GOOS == "plan9" {
+		cmd = "/bin/ape/diff"
+	}
+
+	data, err := exec.Command(cmd, "-u", f1, f2).CombinedOutput()
+	if len(data) > 0 {
+		// diff exits with a non-zero status when the files don't match.
+		// Ignore that failure as long as we get output.
```
amckinney: Are the other potential failures? Should we explicitly check for the error status when the files don't match and otherwise propagate that failure?

internal/diff/diff.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	// Preserve timestamps.
+	var t0, t1 []byte
+	if i := bytes.LastIndexByte(bs[0], '\t'); i != -1 {
+		t0 = bs[0][i:]
+	}
+	if i := bytes.LastIndexByte(bs[1], '\t'); i != -1 {
+		t1 = bs[1][i:]
+	}
+	// Always print filepath with slash separator.
```
amckinney: Is this necessary? As far as I understand it, this will simply map `:` to `/` when the OS uses `:` for its separators. Do we not want to leave it as `:` in those cases?

peter-edge: This one I don't want to adjust as it was copy/pasted from golang/go

internal/diff/diff.go
peter-edge: Oh wait, I indented the wrong thing, lol

internal/diff/diff.go
```diff
@@ -86,7 +86,7 @@ func writeTempFile(dir, prefix string, data []byte) (string, error) {
 //   ...
 //   ->
 //   --- path/to/file.go.orig	2017-02-03 19:13:00.280468375 -0500
-//   +++ path/to/file.go	2017-02-03 19:13:00.280468375 -0500
+//   +++ path/to/file.go		2017-02-03 19:13:00.280468375 -0500
```
amckinney: Looks like this was done one-too-many times :P

PR #53, opened 2018-05-02
Add harbormaster output option

peter-edge:
Per a conversation I had with @AlexAPB. We need a way to produce Harbormaster compatible output, and it seems overkill to create a new wrapper binary for this, especially as Harbormaster is OSS. This adds a flag `--harbormaster` that will output Harbormaster compatible JSON for failures instead of the standard `filename:line:column:message`. Alex, see if this looks good and if you want alternative naming, or if this is not a path you'd want to pursue.

https://secure.phabricator.com/conduit/method/harbormaster.sendmessage

### DIFF
--- a/internal/x/cmd/cmd.go
+++ b/internal/x/cmd/cmd.go
@@ -324,7 +324,7 @@ func getRootCommand(exitCodeAddr *int, args []string, stdin io.Reader, stdout io
 	flags.bindCachePath(rootCmd.PersistentFlags())
 	flags.bindProtocURL(rootCmd.PersistentFlags())
 	flags.bindPrintFields(rootCmd.PersistentFlags())
-	flags.bindHarbormasterOutput(rootCmd.PersistentFlags())
+	flags.bindHarbormaster(rootCmd.PersistentFlags())
 
 	rootCmd.SetArgs(args)
 	rootCmd.SetOutput(stdout)
@@ -384,10 +384,10 @@ func getRunner(stdin io.Reader, stdout io.Writer, stderr io.Writer, flags *flags
 			exec.RunnerWithDirMode(),
 		)
 	}
-	if flags.harbormasterOutput {
+	if flags.harbormaster {
 		runnerOptions = append(
 			runnerOptions,
-			exec.RunnerWithHarbormasterOutput(),
+			exec.RunnerWithHarbormaster(),
 		)
 	}
 	workDirPath, err := os.Getwd()
@@ -424,23 +424,23 @@ func printAndGetErrorExitCode(err error, stdout io.Writer) int {
 }
 
 type flags struct {
-	debug              bool
-	cachePath          string
-	protocURL          string
-	printFields        string
-	dirMode            bool
-	overwrite          bool
-	diffMode           bool
-	lintMode           bool
-	disableFormat      bool
-	disableLint        bool
-	gen                bool
-	headers            []string
-	callTimeout        string
-	connectTimeout     string
-	keepaliveTime      string
-	uncomment          bool
-	harbormasterOutput bool
+	debug          bool
+	cachePath      string
+	protocURL      string
+	printFields    string
+	dirMode        bool
+	overwrite      bool
+	diffMode       bool
+	lintMode       bool
+	disableFormat  bool
+	disableLint    bool
+	gen            bool
+	headers        []string
+	callTimeout    string
+	connectTimeout string
+	keepaliveTime  string
+	uncomment      bool
+	harbormaster   bool
 }
 
 func (f *flags) bindDebug(flagSet *pflag.FlagSet) {
@@ -507,6 +507,6 @@ func (f *flags) bindUncomment(flagSet *pflag.FlagSet) {
 	flagSet.BoolVar(&f.uncomment, "uncomment", false, "Uncomment the example config settings.")
 }
 
-func (f *flags) bindHarbormasterOutput(flagSet *pflag.FlagSet) {
-	flagSet.BoolVar(&f.harbormasterOutput, "harbormaster-output", false, "Print failures in JSON compatible with the Harbormaster API.")
+func (f *flags) bindHarbormaster(flagSet *pflag.FlagSet) {
+	flagSet.BoolVar(&f.harbormaster, "harbormaster", false, "Print failures in JSON compatible with the Harbormaster API.")
 }
--- a/internal/x/cmd/cmd.go
+++ b/internal/x/cmd/cmd.go
@@ -324,6 +324,7 @@ func getRootCommand(exitCodeAddr *int, args []string, stdin io.Reader, stdout io
 	flags.bindCachePath(rootCmd.PersistentFlags())
 	flags.bindProtocURL(rootCmd.PersistentFlags())
 	flags.bindPrintFields(rootCmd.PersistentFlags())
+	flags.bindHarbormasterOutput(rootCmd.PersistentFlags())
 
 	rootCmd.SetArgs(args)
 	rootCmd.SetOutput(stdout)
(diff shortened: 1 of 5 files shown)


internal/x/exec/runner.go
```diff
@@ -811,7 +814,15 @@ func (r *runner) printFailures(filename string, meta *meta, failures ...*text.Fa
 			}
 		}
 		if shouldPrint {
-			if err := failure.Fprintln(bufWriter, failureFields...); err != nil {
+			if r.harbormasterOutput {
+				data, err := json.Marshal(phab.TextFailureToHarbormasterLintResult(failure))
+				if err != nil {
+					return err
```
amckinney: Should we add more context to the error here? It might help to include something about harbormaster so that it's not just the raw json marshal error.

internal/x/phab/phab.go
```diff
@@ -0,0 +1,79 @@
...
+//
+// The primary purpose is to convert failures to JSON that is compatible
+// with the Harbormaster API.
+//
+// https://secure.phabricator.com/conduit/method/harbormaster.sendmessage
+package phab
+
+import "github.com/uber/prototool/internal/x/text"
+
+const (
```
amckinney: Should we un-export these?

internal/x/exec/runner.go
peter-edge: My general philosophy is to not over-complicate (in my opinion) errors here - this will virtually never actually produce an error as it is, and if it does, it’s a system bug that will be easy to find from the error message that is produced. Users can’t take any action on it.

Open to changing.

internal/x/phab/phab.go
```diff
@@ -0,0 +1,79 @@
...
+// TextFailureToHarbormasterLintResult converts a text.Failure to a HarbormasterLintResult.
+func TextFailureToHarbormasterLintResult(textFailure *text.Failure) *HarbormasterLintResult {
+	if textFailure == nil {
+		return nil
+	}
+	harbormasterLintResult := &HarbormasterLintResult{
+		Name:     DefaultHarbormasterLintResultName,
+		Code:     textFailure.ID,
+		Severity: DefaultHarbormasterLintResultSeverity,
+		// TODO: Filename is not a required field on text.Failure
```
amckinney: Should we return an error if the `text.Failure` doesn't have a Filename? If we NEED the Path, it might be best to handle that here.

internal/x/phab/phab.go
peter-edge: I think they should be constant and in the GoDoc (as other packages in Prototool do for now). Note everything will stay internal as it is. But just my opinion.

internal/x/phab/phab.go
peter-edge: Yea fair point. Will update.

internal/x/phab/phab.go
AlexAPB: I'm in favour of keeping them public for godocs since this is in an `internal` package. This allows users to see how we're handling Harbormaster output without trial/error.

internal/x/phab/phab.go
AlexAPB: +1

Meta: For context, why are filenames *not* required for `text.Failure`s? Shouldn't every lint failure should have a corresponding file?

internal/x/phab/phab.go
```diff
@@ -0,0 +1,79 @@
...
+		// but Path is a required field for a Harbormaster Lint Result
+		// There are no good options here
+		Path:        textFailure.Filename,
+		Line:        textFailure.Line,
+		Char:        textFailure.Column,
+		Description: textFailure.Message,
+	}
+	if harbormasterLintResult.Code == "" {
```
AlexAPB: Meta: Would this case realistically ever happen since `textFailure.ID` corresponds to the name of the linters?

AlexAPB: Actually, assuming this case rarely happens, should we make `DefaultHarbormasterLintResultCode` private?

If we ever set this error code with that default value, it seems like a linter would have been improperly set up. Then, `DefaultHarbormasterLintResultCode` just becomes documentation noise.

internal/x/cmd/cmd.go
```diff
@@ -498,3 +506,7 @@ func (f *flags) bindKeepaliveTime(flagSet *pflag.FlagSet) {
 func (f *flags) bindUncomment(flagSet *pflag.FlagSet) {
 	flagSet.BoolVar(&f.uncomment, "uncomment", false, "Uncomment the example config settings.")
 }
+
+func (f *flags) bindHarbormasterOutput(flagSet *pflag.FlagSet) {
+	flagSet.BoolVar(&f.harbormasterOutput, "harbormaster-output", false, "Print failures in JSON compatible with the Harbormaster API.")
```
AlexAPB: maybe just `harbormaster`? @amckinney thoughts?

AlexAPB:
I'm fine with this approach for now to unblock us, but I think we should leave it in `/x`. This lets us circle back if we later decide this isn't the right approach.

internal/x/phab/phab.go
peter-edge: I’d vote for this too. Note that this can change later - let’s keep it in x/, the goal here really is to make sure we’re all “generally” on the same page so I can trial this internally.

@amckinney is this OK? Up to you two really.

internal/x/phab/phab.go
peter-edge: protoc parsing can be an issue, and sometimes you don’t get a filename, see the regex parsing in the protoc package. I’m going to think more about this tomorrow and see if I can come up with better guarantees.

internal/x/phab/phab.go
peter-edge: So (not that it needs to stay this way, just an FYI) I generally make such “informational” constants public, so they are documented and so they are forced to stay the same - whether private or public, changing these is a breaking change technically, so might as well be public to hit the point home.

internal/x/cmd/cmd.go
peter-edge: That was my first thought, happy to change to that.

internal/x/phab/phab.go
peter-edge: One thing I might do is fail the prototool call with a system error if harbormaster output can not be created for this reason. So high level, if you specify the harbormaster flag, and a failure does not have a filename, you won’t get a text.Failure, you’ll get a system error at the top level. Thoughts?

AlexAPB: sounds good to me!

internal/x/phab/phab.go
AlexAPB: Makes sense. 
Can you add to the documentation of `DefaultHarbormasterLintResultCode` that the default will only be used if we don't have a linter name?

internal/x/phab/phab.go
amckinney: Fine with me! Just thought it was worth mentioning. This makes sense with respect to a godoc perspective.

internal/x/exec/runner.go
amckinney: Good point, I'm fine with leaving this as is. If it becomes an issue later on (doubtful it will), it's an easy change.

internal/x/phab/phab.go
amckinney: Nice, sounds good 👍

internal/x/cmd/cmd.go
amckinney: Yea, let's shorten to `harbormaster`.

amckinney:
I agree with @AlexAPB that keeping this in `x/` is best. Once the system-level error for missing filenames is in, this looks good to me.

internal/x/phab/phab.go
peter-edge: Done

peter-edge:
I addressed the comments, and return a system error if `Filename` is not set on a `text.Failure`. Will merge, but staying in x/ for now.

internal/x/cmd/cmd.go
peter-edge: Done

internal/x/phab/phab.go
peter-edge: Done

PR #54, opened 2018-05-02
fix a typo in the example

termie:
trivial typo, will go figure out where it is done in the proto init and change it there too

Thanks for your pull request! We really appreciate all OSS efforts, big or small.

Before submitting, make sure to:

- Run `make init` if you have not already to download dependencies to `vendor`.
- Run `make generate` to make sure there is no diff.
- Run `make` to make sure all tests pass. This is functionally equivalent to the tests run on CI.


termie:
making a new patch fixing all in one place

PR #55, opened 2018-05-02
fix typo in cfginit

termie:
Thanks for your pull request! We really appreciate all OSS efforts, big or small.

Before submitting, make sure to:

- Run `make init` if you have not already to download dependencies to `vendor`.
- Run `make generate` to make sure there is no diff.
- Run `make` to make sure all tests pass. This is functionally equivalent to the tests run on CI.


termie:
this is comparing against a golden file, i'll just make a new patch fixing all these in one

Issue #59, opened 2018-05-04
Handle repeated options


peter-edge:
This came up when fixing https://github.com/uber/prototool/issues/27. Will fix in a follow-up PR.

peter-edge:
Fixed.

PR #60, opened 2018-05-04
Handle nested options in the formatter

### LINKED ISSUE #27: Formatting nested options produces code that emicklei/proto cannot parse
When the formatter processes the following file:

```protobuf
syntax = "proto3";

import "google/api/annotations.proto";

service Test {
    rpc Test(Void) returns (Void) {
        option (google.api.http) = {
            get: "/api/v1/test"
            additional_bindings {
                post: "/api/v1/test"
                body: "*"
            }
        };
    }
}

message Void {}
```
It produces the following:
```protobuf
syntax = "proto3";

import "google/api/annotations.proto";

service Test {
	rpc Test(Void) returns (Void) {
		option (google.api.http) = {
			get: "/api/v1/test"
			additional_bindings.post: "/api/v1/test"
			additional_bindings.body: "*"
		};
	}
}

message Void {}
```

This output is not parsable by `emicklei/proto` at the moment (see https://github.com/emicklei/proto/issues/80).

The options as I see it:
1. Change formatter to output the same code as in sample 1 above.
2. Wait for https://github.com/emicklei/proto/issues/80 to be fixed.
3. Replace `emicklei/proto` with something else (see #1).

peter-edge:
After a fix done in emicklei/proto, we can now handle nested options. This fixes #27.

Message options with a single value are no longer special-cased, it's more difficult now that there is a recursive nature to the printing of option values.

As you can see in the code, there's some weirdness in emicklei/proto regarding options printing, this is a good example of why #1 should maybe be done.

### DIFF
--- a/internal/x/format/base_visitor.go
+++ b/internal/x/format/base_visitor.go
@@ -138,7 +138,7 @@ func (v *baseVisitor) POptions(isFieldOption bool, options ...*proto.Option) {
 		if len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 {
 			v.PWithInlineComment(o.InlineComment, prefix, o.Name, ` = `, o.Constant.SourceRepresentation(), suffix)
 		} else if len(o.Constant.Array) > 0 { // both Array and OrderedMap should not be set simultaneously, need more followup with emicklei/proto
-			// TODO
+			// TODO https://github.com/uber/prototool/issues/59
 		} else { // len(o.Constant.OrderedMap) > 0
 			v.P(prefix, o.Name, ` = {`)
 			v.In()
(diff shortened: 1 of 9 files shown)


internal/x/format/base_visitor.go
```diff
@@ -134,23 +134,45 @@ func (v *baseVisitor) POptions(isFieldOption bool, options ...*proto.Option) {
...
+// should only be called by Poptions
+func (v *baseVisitor) pInnerLiteral(name string, literal proto.Literal) {
+	prefix := ""
+	if name != "" {
+		prefix = name + ": "
+	}
+	// TODO: this is a good example of the reasoning for https://github.com/uber/prototool/issues/1
+	if len(literal.Array) == 0 && len(literal.OrderedMap) == 0 {
+		v.P(prefix, literal.SourceRepresentation())
+	} else if len(literal.Array) > 0 { // both Array and OrderedMap should not be set simultaneously, need more followup with emicklei/proto
```
AlexAPB: Clarification: will options ever actually be filled within `Literal.Array`? From what I can tell (though probably wrong) I think that this should always be empty when dealing with options.

internal/x/format/base_visitor.go
```diff
@@ -134,23 +134,45 @@ func (v *baseVisitor) POptions(isFieldOption bool, options ...*proto.Option) {
...
-		switch len(o.AggregatedConstants) {
-		case 0:
+		// TODO: this is a good example of the reasoning for https://github.com/uber/prototool/issues/1
+		if len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 {
 			v.PWithInlineComment(o.InlineComment, prefix, o.Name, ` = `, o.Constant.SourceRepresentation(), suffix)
-		case 1:
-			v.PWithInlineComment(o.InlineComment, prefix, o.Name, ` = { `, o.AggregatedConstants[0].Name, ": ", o.AggregatedConstants[0].Literal.SourceRepresentation(), " }", suffix)
-		default:
+		} else if len(o.Constant.Array) > 0 { // both Array and OrderedMap should not be set simultaneously, need more followup with emicklei/proto
+			// TODO https://github.com/uber/prototool/issues/59
```
AlexAPB: Should we put a log statement here? or would this make too much noise?

internal/x/format/base_visitor.go
peter-edge: I need to address this in a follow-up PR/needs more investigation. This PR is fixing a bug, basically. I filed https://github.com/uber/prototool/issues/59 to track this, this will come next.

internal/x/format/base_visitor.go
peter-edge: Will test out in a follow-up PR. Basically need to test how to parse:

```
extend google.protobuf.FieldOptions {
  repeated int64 custom_repeated_option = 80001;
}
```

I think it will be:

```
message Foo {
  string bar = 1 [(custom_repeated_option) = [1, 2, 3]];
}
```

But I am not sure.

Would not want to do logging personally.

peter-edge:
Merging this now as this is a semi-important bug fix. Will work on https://github.com/uber/prototool/issues/59 this week.

Issue #63, opened 2018-05-06
Lint failure on protocol buffers version 2.6.1


nikostrongioglou:
I am using protocol buffers version 2.6.1 and linting fails with the following error on debug

```
2018-05-06T23:59:52.701+0200	DEBUG	downloaded protobuf zip file	{"url": "https://github.com/google/protobuf/releases/download/v2.6.1/protoc-2.6.1-osx-x86_64.zip"}
zip: not a valid zip file
```
The URL indeed does not work and as you can see from the file `prototool` tries to download we are talking about an OSX system here.

Is `prototool` compatible with version 2 protos?

PR #66, opened 2018-05-07
Add list of valid protoc versions from GitHub Releases

peter-edge:
See #63. This adds a `map[string]struct{}` of protoc versions that have a valid `protoc-VERSION-OS-ARCH.zip` uploaded to GitHub Releases (for example, https://github.com/google/protobuf/releases/v3.2.0), and returns an error if a protoc version is not known.

### DIFF
--- a/internal/x/protoc/downloader.go
+++ b/internal/x/protoc/downloader.go
@@ -221,6 +221,11 @@ func (d *downloader) getProtocURL(goos string, goarch string) (string, error) {
 	if d.protocURL != "" {
 		return d.protocURL, nil
 	}
+	// if protocURL is not set, we need to make sure we have a version that will result
+	// in a GitHub Releases link with a protoc zip file
+	if _, ok := vars.ValidProtocVersions[d.config.Compile.ProtobufVersion]; !ok {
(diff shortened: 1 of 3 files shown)


internal/x/protoc/downloader.go
```diff
@@ -221,6 +221,11 @@ func (d *downloader) getProtocURL(goos string, goarch string) (string, error) {
 	if d.protocURL != "" {
 		return d.protocURL, nil
 	}
+	// if protocURL is not set, we need to make sure we have a version that will result
+	// in a GitHub Releases link with a protoc zip file
+	if _, ok := vars.ValidProtocVersions[d.config.Compile.ProtobufVersion]; !ok {
+		return "", fmt.Errorf("%s is not a known to have a protoc zip file in the proper format downloadable from https://github.com/google/protobuf/releases/v%s", d.config.Compile.ProtobufVersion, d.config.Compile.ProtobufVersion)
```
AlexAPB: typo: `is not known`

AlexAPB:
Would a good middle ground be annotating download errors with a verbose message?
 eg:
`... Ensure that this version of protoc has the corresponding zip package by visiting github.com/google/protobuf/releases.`

cc @amckinney

amckinney:
I agree - using a verbose message here is a great idea.

AlexAPB:
closing since we're all in agreement

PR #68, opened 2018-05-07
Handle repeated option values in the formatter

peter-edge:
See https://github.com/uber/prototool/issues/59.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,7 +1,7 @@
 package: github.com/uber/prototool
 import:
   - package: github.com/emicklei/proto
-    version: 0cbee195654dd0aab66561d51e08d37554fd8cfe
+    version: 43f8fb7abe35485d1ca39fe380c5b6d39aa931c2
   - package: github.com/fullstorydev/grpcurl
   - package: github.com/gogo/protobuf/protoc-gen-gogoslick
   - package: github.com/golang/protobuf/protoc-gen-go
--- a/internal/x/format/base_visitor.go
+++ b/internal/x/format/base_visitor.go
@@ -138,12 +138,15 @@ func (v *baseVisitor) POptions(isFieldOption bool, options ...*proto.Option) {
 		if len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 {
 			v.PWithInlineComment(o.InlineComment, prefix, o.Name, ` = `, o.Constant.SourceRepresentation(), suffix)
 		} else if len(o.Constant.Array) > 0 { // both Array and OrderedMap should not be set simultaneously, need more followup with emicklei/proto
-			// TODO https://github.com/uber/prototool/issues/59
+			v.Failures = append(
+				v.Failures,
+				text.NewFailuref(o.Position, "INVALID_PROTOBUF", "top-level options should never be arrays, this should not compile with protoc"),
+			)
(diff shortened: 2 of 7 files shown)


glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/uber/prototool
 import:
   - package: github.com/emicklei/proto
-    version: 0cbee195654dd0aab66561d51e08d37554fd8cfe
+    version: 43f8fb7abe35485d1ca39fe380c5b6d39aa931c2
```
amckinney: It's unfortunate that we need to pin to a direct hash here. We should try and get SemVer tags to depend on sometime in the future.

peter-edge: Yea, but it's just how it is for now - emicklei/proto makes breaking changes, even though it does releases. See #1.

There's no real good alternative to this for now.

internal/x/format/base_visitor.go
```diff
@@ -138,12 +138,15 @@ func (v *baseVisitor) POptions(isFieldOption bool, options ...*proto.Option) {
 		if len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 {
 			v.PWithInlineComment(o.InlineComment, prefix, o.Name, ` = `, o.Constant.SourceRepresentation(), suffix)
 		} else if len(o.Constant.Array) > 0 { // both Array and OrderedMap should not be set simultaneously, need more followup with emicklei/proto
-			// TODO https://github.com/uber/prototool/issues/59
+			v.Failures = append(
+				v.Failures,
+				text.NewFailuref(o.Position, "INVALID_PROTOBUF", "top-level options should never be arrays, this should not compile with protoc"),
```
amckinney: should not or cannot?

peter-edge: Cannot, to the best I can figure out. However, it's a valid "switch" here, so I want to record it as a failure if we somehow hit it, as opposed to just ignoring it.

internal/x/format/base_visitor.go
```diff
@@ -138,12 +138,15 @@ func (v *baseVisitor) POptions(isFieldOption bool, options ...*proto.Option) {
...
+			)
 		} else { // len(o.Constant.OrderedMap) > 0
 			v.P(prefix, o.Name, ` = {`)
 			v.In()
 			for _, namedLiteral := range o.Constant.OrderedMap {
-				v.pInnerLiteral(namedLiteral.Name, *namedLiteral.Literal)
+				v.pInnerLiteral(namedLiteral.Name, *namedLiteral.Literal, "")
```
amckinney: nit: consider `v.pInnerLiteral(namedLiteral.Name, *namedLiteral.Literal, "" /* suffix */)`

peter-edge: Is this necessary here? The function definition is right below 10 lines down, and I think this just muddies the water personally :-)

amckinney: Yea, not necessarily. Since the definition is so local it's probably overkill.

internal/x/format/base_visitor.go
peter-edge: Whoops, already merged. This error message should never print so it shouldn't (can't :-) ) be a big deal, but I can do a subsequent PR if it's a problem.

PR #71, opened 2018-05-09
Move internal/x/cmd to internal/cmd

peter-edge:
Follow-up to #51, with two commits. Flags are also moved to a separate file, and all commands, flag bind calls, and flag accessors are alphabetically ordered per the first comment on #51.


peter-edge:
Don't worry about the conflicting files, they are easily merged and don't affect this review.

peter-edge:
Did a merge instead of rebase as the rebase got really confusing with the delete/add.

peter-edge:
https://github.com/uber/prototool/pull/71/commits/6fbe3bdd07ec4f325b26837779749e20cc3b0d4b is the commit where the files were re-added and is what you want to review assumedly.

internal/cmd/cmd.go
```diff
@@ -236,95 +257,81 @@ func getRootCommand(exitCodeAddr *int, args []string, stdin io.Reader, stdout io
...
+	rootCmd.AddCommand(versionCmd)
 
-	flags.bindDebug(rootCmd.PersistentFlags())
 	flags.bindCachePath(rootCmd.PersistentFlags())
-	flags.bindProtocURL(rootCmd.PersistentFlags())
-	flags.bindPrintFields(rootCmd.PersistentFlags())
+	flags.bindDebug(rootCmd.PersistentFlags())
 	flags.bindHarbormaster(rootCmd.PersistentFlags())
+	flags.bindPrintFields(rootCmd.PersistentFlags())
+	flags.bindProtocURL(rootCmd.PersistentFlags())
```
amckinney: It would help to document how the flags are persisted somewhere. It doesn't necessarily need to be here (and maybe it already exists somewhere), but I think this might help with potential debugging later on. As far as I understand it, the "debug" flag is bound on the `rootCmd` and is therefore passed along to all of its sub-commands (everything in this case).

peter-edge: What would you propose? I can put a comment that all flags put on `rootCmd` are propagated to all sub-commands, does that work? The propagation is part of the Cobra API too.

amckinney: Yea that's fair. I guess this was mostly just a foreign library use to me and required some digging around. I think it's justifiable to leave it out for now.

Issue #69, opened 2018-05-08
swagger annotations break linter


lucasvo:
I'm trying to get the linter to pass on my swagger annotations however it seems like protoc is invoked differently for `prototool lint` command than for `prototool gen`. The modifications below to the example break lint but work for gen.  Is this an issue with my configuration or are the swagger annotations not supported yet?

```bash,
# prototool lint
[...]
2018-05-08T12:07:26.706-0400	DEBUG	running protoc	{"command": "/Users/lucasvo/Library/Caches/prototool/Darwin/x86_64/protobuf/3.5.1/bin/protoc -I /Users/lucasvo/Code/go/src/github.com/uber/prototool/vendor/github.com/grpc-ecosystem/grpc-gateway -I /Users/lucasvo/Code/go/src/github.com/uber/prototool/vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I /Users/lucasvo/Library/Caches/prototool/Darwin/x86_64/protobuf/3.5.1/include -I /Users/lucasvo/Code/go/src/github.com/uber/prototool/example/idl/uber -o /dev/null /Users/lucasvo/Code/go/src/github.com/uber/prototool/example/idl/uber/foo/foo.proto"}
foo/foo.proto:18:1: found "}" but expected [.proto element {comment|option|import|syntax|enum|service|package|message}]

# prototool gen
[...]
[...]

PR #72, opened 2018-05-09
Add extra logging

peter-edge:
Per #69, extra logging.


internal/x/exec/runner.go
```diff
@@ -340,6 +341,7 @@ func (r *runner) Lint(args []string) error {
 }
 
 func (r *runner) lint(meta *meta) error {
+	r.logger.Debug("calling LintRunner")
```
AlexAPB: just a nit: since this is exposed to cli users who don't know what a `lintRunner` is, should we just have this as `"calling lint"`?

peter-edge: Debug lines expose a lot of internals on purpose, see the logging lines in the file and settings packages for example.

AlexAPB: 👍

PR #73, opened 2018-05-10
Fixed spelling for StreamQuasars

beono:
Fixed spelling for `StreamQuasars`.
Also remove `Quadrant` from `StreamQuasarsRequest`

### DIFF
--- a/etc/style/uber/uber.proto
+++ b/etc/style/uber/uber.proto
@@ -208,12 +208,6 @@ message Galaxy {
   string name = 2;
 }
 
-// Quadrant is a quadrant.
-message Quadrant {
-  string id = 1;
-  string name = 2;
-}
-
 // Quasar is a quasar.
 message Quasar {
   string id = 1;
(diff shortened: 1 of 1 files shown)


etc/style/uber/uber.proto
```diff
@@ -265,11 +259,11 @@ message StreamGalaxiesResponse {
   Galaxy galaxy = 1;
 }
 
-message StreamQuasarsRequest {
-  Quadrant quadrant = 1;
+message GetQuasarsRequest {
```
peter-edge: Thanks for the PR!

Why do this? We wanted this named `StreamQuasars`.

I can merge this if you want to remove `Quadrant` and fix the spelling of `strams`, but want to keep this.

beono: @peter-edge 'cause it's not a stream.
As I understand prefix `Stream` should be used only when it returns a stream.

Consider this example

```
service HelloService {
  rpc StreamQuasars(stream GetQuasarsRequest) returns (stream StreamQuasarsResponse) {}
  rpc GetQuasars(stream GetQuasarsRequest) returns (GetQuasarsResponse) {}
}
```

beono: @peter-edge Ok, I will update my pr. But it's not clear when should I use `Stream` prefix

peter-edge: There’s no naming convention in this style guide regarding streams. One of the methods is a client stream, the other a server stream, and the point being to end requests and responses with the same postfix as unary methods.

If this is reverted, I can merge.

peter-edge: There’s no style recommendation on prefixes.

beono:
@peter-edge I've updated the PR

PR #74, opened 2018-05-14
Add more integration testing to internal/cmd

peter-edge:
This chains on top of #71 to add more integration testing.


internal/cmd/cmd_test.go
```diff
@@ -354,6 +580,20 @@ func assertGRPC(t *testing.T, expectedExitCode int, expectedLinePrefixes string,
 	assertDoStdin(t, strings.NewReader(jsonData), expectedExitCode, expectedLinePrefixes, "grpc", filePath, excitedTestCase.Address(), method, "-")
 }
 
+func assertRegexp(t *testing.T, expectedExitCode int, expectedRegexp string, args ...string) {
+	stdout, exitCode := testDo(t, args...)
+	assert.Equal(t, expectedExitCode, exitCode)
+	matched, err := regexp.MatchString(expectedRegexp, stdout)
+	assert.NoError(t, err)
+	assert.True(t, matched, stdout)
```
amckinney: Do you think it's overkill to include the `expectedRegexp` in the `assert.True` failure message?

peter-edge: Yea I can do that, give me a few mins

amckinney: np - this might also apply to the other test helpers.

amckinney:
How do we use the commented out `-debug` flag? Is that a contributor's tool for diagnosing test failures? If so, I wonder if there's a better way to handle that behavior, such as in a build rule. I'm not sure exactly how that would look, but it might be worth exploring.

internal/cmd/cmd_test.go
peter-edge: The other helper I see right now prints the output as the message, which is basically what I want, is that OK?

peter-edge:
If anything I can just delete the commented-out `--debug` flag.

peter-edge:
I like it there just because I use it occasionally during development, it's easier to uncomment than to have an env var or something.

amckinney:
Ok, seeing as you haven't used it in a while, let's leave it out for now. We can always reintroduce something similar if there becomes a large need for it.

peter-edge:
Done

amckinney:
And with respect to the `testCleaned` flag - 

Do we only ever want to "clean" a test environment once? I notice that this call happens in `testDownload` which takes place in both `testDo` and `testDoStdin` which are called in a variety of places. Should we ever reset it to `false`?

PR #75, opened 2018-05-14
Add all lint group

peter-edge:
This adds an `all` lint group, that contains all known linters. The default set of linters does not include all the linters, as some were removed, but users may want them in the future. The `all` group is primarily for testing, and testing will be added in #74.

### DIFF
--- a/internal/x/exec/runner.go
+++ b/internal/x/exec/runner.go
@@ -380,7 +380,12 @@ func (r *runner) ListLintGroup(group string) error {
 }
 
 func (r *runner) ListAllLintGroups() error {
+	groups := make([]string, 0, len(lint.GroupToCheckers))
 	for group := range lint.GroupToCheckers {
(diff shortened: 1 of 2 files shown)


internal/x/exec/runner.go
```diff
@@ -380,7 +380,12 @@ func (r *runner) ListLintGroup(group string) error {
 }
 
 func (r *runner) ListAllLintGroups() error {
+	groups := make([]string, 0, len(lint.GroupToCheckers))
 	for group := range lint.GroupToCheckers {
+		groups = append(groups, group)
+	}
+	sort.Strings(groups)
```
amckinney: Adding this for determinism?

peter-edge: Yea, cuz I'll be testing this in #74 and you want the output to be deterministic. Even without wanting this for testing, it's just nice to have the output sorted.

PR #77, opened 2018-05-14
Move internal/x/settings to internal/settings

peter-edge:


internal/settings/config_provider.go
```diff
@@ -0,0 +1,399 @@
...
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
+)
+
+type configProvider struct {
+	logger                   *zap.Logger
+	filePathToConfig         map[string]Config
+	dirPathToFilePath        map[string]string
+	dirPathToExcludePrefixes map[string][]string
+	lock                     sync.RWMutex
```
amckinney: There is a lot of logic written around making the `configProvider` concurrent-safe. Is this really necessary? In the places that this might be called concurrently, such as in calls in `internal/x/file`, can we instead synchronously run the commands? 

I'm not sure how much of a performance gain we expect to see in this domain, so it might be worth removing it for now. 

What do you think? Is this actually completely necessary / am I missing something?

peter-edge: Things like caches (which is what this is locking) should always be thread-safe, IMO, I don't see it as an optional thing. I don't want our designs to have to rely on something being not thread-safe, or worse, someone using this in a deep dependency in a threaded manner without realizing this is not thread-safe. It's defensive programming if anything, but a cache should generally always be thread-safe unless there's a compelling reason not to be IMO.

peter-edge:
I ripped out the caching in the last diff - it actually doesn't clean it up that much, but it doesn't seem to make any difference in time when I run it against googleapis. I'll play with it more but do you like this more?

internal/settings/settings.go
```diff
@@ -0,0 +1,350 @@
...
+	RelPath string
+	AbsPath string
+}
+
+// ExternalConfig is the external representation of Config.
+//
+// It is meant to be set by a YAML or JSON config file, or flags.
+type ExternalConfig struct {
+	Excludes           []string `json:"excludes,omitempty" yaml:"excludes,omitempty"`
+	NoDefaultExcludes  bool     `json:"no_default_excludes,omitempty" yaml:"no_default_excludes,omitempty"`
```
amckinney: Re: https://github.com/uber/prototool/pull/49 - 

Should we remove this?

internal/settings/config_provider.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		return Config{}, err
+	}
+	includePaths := make([]string, 0, len(e.ProtocIncludes))
+	for _, includePath := range strs.DedupeSortSlice(e.ProtocIncludes, nil) {
+		if !filepath.IsAbs(includePath) {
+			includePath = filepath.Join(dirPath, includePath)
+		}
+		includePath = filepath.Clean(includePath)
+		//if includePath != dirPath {
```
amckinney: Should we remove this?

amckinney:
I prefer the implementation without the cache, especially because I don't know how often we need to retrieve a configuration for the same file path (rather than simply accessing the configuration that was already retrieved once). It does, however, make it so that we will unmarshal the yaml every time the file path is requested.

I'm not saying that we should do this here, but we should consider the `prototool.yaml` hierarchy and determine whether or not we should stick to the current tree/override behavior. More details https://github.com/uber/prototool/issues/93

internal/settings/config_provider.go
```diff
@@ -28,27 +28,19 @@ import (
...
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/uber/prototool/internal/strs"
 	"go.uber.org/zap"
 	"gopkg.in/yaml.v2"
 )
 
 type configProvider struct {
```
amckinney: nit: Now that the cache is removed, it looks like we don't actually need this type. The `logger` is never used and all of the `configProvider` methods don't access any of the type's attributes. In other words, all of the methods could be package-level functions now. 

Removing this later might become more relevant depending on the direction we take https://github.com/uber/prototool/issues/93. If we only ever need one `prototool.yaml`, simple package-level functions might make more sense since we only need the single configuration struct.

peter-edge: I still want to keep this abstraction for now, but we can revisit after #93.

Issue #78, opened 2018-05-14
Formatter omits empty option incorrectly


johanbrandhorst:
When running `prototool format` on the following proto:

```protobuf
syntax = "proto3";

import "protoc-gen-swagger/options/annotations.proto";

option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
	info: {
		version: "1.0"
	}
	external_docs: {}
	schemes: HTTPS
};
```

It produces the following output:

```protobuf
syntax = "proto3";

import "protoc-gen-swagger/options/annotations.proto";

option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
	info: {
		version: "1.0"
	}
	external_docs:
	schemes: HTTPS
};
```

This does not compile with `protoc`, instead giving the error:

```bash
$  
test.proto: Error while parsing option value for "openapiv2_swagger": Expected "{", found "schemes".
```

The correct thing here would be to leave the `{}`.

peter-edge:
Question: why does the key `external_docs` need to be outputted? Is it proto2 and required? Why is it correct to output it from the formatter, as opposed to drop it?

Figuring out the type of the value for `external_docs` and outputting `0, [], {}`, etc will be difficult.

johanbrandhorst:
Well, it can be dropped, but if so the key should be dropped too. As it is it drops the value but not the key.

PR #79, opened 2018-05-15
Do not print option keys with empty values

peter-edge:
Fixes https://github.com/uber/prototool/issues/78.


internal/x/format/base_visitor.go
```diff
@@ -162,7 +164,9 @@ func (v *baseVisitor) pInnerLiteral(name string, literal proto.Literal, suffix s
 	}
 	// TODO: this is a good example of the reasoning for https://github.com/uber/prototool/issues/1
 	if len(literal.Array) == 0 && len(literal.OrderedMap) == 0 {
-		v.P(prefix, literal.SourceRepresentation(), suffix)
+		if source := literal.SourceRepresentation(); source != "" {
```
amckinney: It's pretty unclear what this branch is checking / doing. Could we add some more context here, perhaps with a new helper function or in-line description?

peter-edge: Can I just add a comment that says “SourceReprrsentation() returns an empty string if the literal is empty, and if empty, we do not want to print the key or empty value”?

PR #81, opened 2018-05-15
Simplify internal/strs

amckinney:
As discussed in https://github.com/uber/prototool/pull/50, this makes several changes to the `internal/strs` package in an attempt at simplifying.

A couple functions using the `containsRune` helper (via the `extraRunes ...rune` parameter) were not actually being used, so it seems better to remove for now. The package name linter used this by allowing `.` characters, so we explicitly handle that there.

Given that the [Protobuf spec](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec) strictly specifies valid identifiers as `A-Z`, `a-z` and `_`, this does not use the corresponding `unicode` functions, such as `unicode.IsLetter`. Instead, we explicitly check for the valid identifiers in a few helper functions, `isUpper`, `isLower`, and `isDigit` so that we restrict the valid input to what is expected by Proto. This is functionally the same as before.

### DIFF
--- a/internal/strs/strs.go
+++ b/internal/strs/strs.go
@@ -90,23 +90,9 @@ func ToUpperSnakeCase(s string) string {
 	return strings.ToUpper(toSnake(s))
 }
 
-// toSnake converts s to snake_case.
-// It is assumed s has no spaces.
-func toSnake(s string) string {
-	output := ""
-	for i, c := range s {
-		if i > 0 && unicode.IsUpper(c) && output[len(output)-1] != '_' && i < len(s)-1 && !unicode.IsUpper(rune(s[i+1])) {
-			output += "_" + string(c)
-		} else {
-			output += string(c)
-		}
-	}
-	return output
-}
-
-// Dedupe returns s with no duplicates and no empty strings, sorted.
+// DedupeSort returns s with no duplicates and no empty strings, sorted.
 // If modifier is not nil, modifier will be applied to each element in s.
-func Dedupe(s []string, modifier func(string) string) []string {
+func DedupeSort(s []string, modifier func(string) string) []string {
 	m := make(map[string]struct{}, len(s))
 	o := make([]string, 0, len(m))
 	for _, e := range s {
@@ -171,6 +157,20 @@ func IsUppercase(s string) bool {
 		return false
 	}
 	return strings.ToUpper(s) == s
+}
+
+// toSnake converts s to snake_case.
+// It is assumed s has no spaces.
+func toSnake(s string) string {
+	output := ""
+	for i, c := range s {
+		if i > 0 && unicode.IsUpper(c) && output[len(output)-1] != '_' && i < len(s)-1 && !unicode.IsUpper(rune(s[i+1])) {
+			output += "_" + string(c)
+		} else {
+			output += string(c)
+		}
+	}
+	return output
 }
 
 func isLetter(r rune) bool {
--- a/internal/strs/strs.go
+++ b/internal/strs/strs.go
@@ -29,99 +29,70 @@ import (
 	"sort"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 )
 
 // IsCapitalized returns true if is not empty and the first letter is
-// between 'A' and 'Z'.
+// an uppercase character.
 func IsCapitalized(s string) bool {
 	if s == "" {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


internal/x/lint/check_package_lower_snake_case.go
```diff
@@ -59,7 +61,7 @@ func (v *packageLowerSnakeCaseVisitor) VisitPackage(pkg *proto.Package) {
 
 func (v *packageLowerSnakeCaseVisitor) Finally() error {
 	if v.pkg != nil {
-		if !strs.IsLowerSnakeCase(v.pkg.Name, '.') {
+		if !strs.IsLowerSnakeCase(strings.Replace(v.pkg.Name, ".", "", -1)) {
```
peter-edge: This is super cute, I love it +1

internal/strs/strs.go
```diff
@@ -133,20 +104,10 @@ func ToSnakeCase(s string) string {
 	return output
 }
 
-// DedupeSlice returns s with no duplicates and no empty strings, in the same order.
+// Dedupe returns s with no duplicates and no empty strings, sorted.
```
peter-edge: nit: I think this should still be `DedupeSort` as a form of self-documentation. If the function is just named `Dedupe`, and I input `[]{"b", "a", "c"}`, my natural instinct would be to expect to get the same slice value back. People may miss the Godoc.

internal/strs/strs_test.go
```diff
@@ -53,8 +52,6 @@ func TestIsLowerSnakeCase(t *testing.T) {
 	assert.False(t, IsLowerSnakeCase("Hello_world"))
 	assert.False(t, IsLowerSnakeCase("_hello_world"))
 	assert.False(t, IsLowerSnakeCase("hello_world_"))
-	assert.False(t, IsLowerSnakeCase("hello_world.foo"))
```
peter-edge: I think this test case should remain.

internal/strs/strs_test.go
```diff
@@ -67,8 +64,6 @@ func TestIsUpperSnakeCase(t *testing.T) {
 	assert.False(t, IsUpperSnakeCase("Hello_world"))
 	assert.False(t, IsUpperSnakeCase("_HELLO_WORLD"))
 	assert.False(t, IsUpperSnakeCase("HELLO_WORLD_"))
-	assert.False(t, IsUpperSnakeCase("HELLO_WORLD.FOO"))
```
peter-edge: I think this test case should remain.

internal/strs/strs.go
```diff
@@ -29,99 +29,70 @@ import (
...
 }
 
-// ToSnakeCase converts s to Snake_case.
+// toSnake converts s to snake_case.
 // It is assumed s has no spaces.
-func ToSnakeCase(s string) string {
+func toSnake(s string) string {
 	output := ""
 	for i, c := range s {
 		if i > 0 && unicode.IsUpper(c) && output[len(output)-1] != '_' && i < len(s)-1 && !unicode.IsUpper(rune(s[i+1])) {
```
peter-edge: We did `unicode.IsUpper` in #50, we should probably change it back.

internal/strs/strs.go
```diff
@@ -29,99 +29,70 @@ import (
...
 func ToUpperSnakeCase(s string) string {
-	return strings.ToUpper(ToSnakeCase(s))
+	return strings.ToUpper(toSnake(s))
 }
 
-// ToSnakeCase converts s to Snake_case.
+// toSnake converts s to snake_case.
 // It is assumed s has no spaces.
-func ToSnakeCase(s string) string {
+func toSnake(s string) string {
```
peter-edge: You'll notice in prototool that the general style is to put public functions at the top, and private functions below, so I would move this to the bottom.

internal/strs/strs.go
amckinney: Good call! Will update.

internal/strs/strs_test.go
amckinney: Yea good call here - removing it was lazy of me.

internal/strs/strs.go
amckinney: Oops, nice catch! Meant to do that here as well.

internal/strs/strs.go
amckinney: Sweet, works for me.

PR #83, opened 2018-05-18
Small refactor in internal/diff

amckinney:
This is a simple refactor for centralizing the `os.Remove` call in `Do` rather than handling it in multiple functions. Note that this is functionally equivalent and not a necessary change.

This also removes the `filepath.ToSlash` call when outputting our diff. We currently only support operating systems that use `/` as a separator, so this is effectively a no-op and should be harmless to remove. If we eventually support other operating systems with another separator, as in `:`, we should probably respect that output and keep the format as they expect it (i.e. don't transform `:` to `/`).

### DIFF
--- a/internal/diff/diff.go
+++ b/internal/diff/diff.go
@@ -31,23 +31,22 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"runtime"
 )
 
 // Do does a diff between an input and output.
 func Do(input []byte, output []byte, filename string) ([]byte, error) {
 	f1, err := writeTempFile("", "prototool-diff", input)
+	defer func() { os.Remove(f1) }()
 	if err != nil {
 		return nil, err
 	}
-	defer func() { _ = os.Remove(f1) }()
 
 	f2, err := writeTempFile("", "prototool-diff", output)
+	defer func() { os.Remove(f2) }()
 	if err != nil {
 		return nil, err
 	}
-	defer func() { _ = os.Remove(f2) }()
 
 	cmd := "diff"
 	if runtime.GOOS == "plan9" {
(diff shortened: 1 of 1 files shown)


internal/diff/diff.go
```diff
@@ -31,23 +31,22 @@ import (
...
 	"os"
 	"os/exec"
-	"path/filepath"
 	"runtime"
 )
 
 // Do does a diff between an input and output.
 func Do(input []byte, output []byte, filename string) ([]byte, error) {
 	f1, err := writeTempFile("", "prototool-diff", input)
+	defer func() { _ = os.Remove(f1) }()
```
peter-edge: Why is this above the error statement? If err is not nil, then f1 could be nil, meaning you're calling `os.Remove` on nil. This should probably be put back to where it was, same below.

internal/diff/diff.go
```diff
@@ -69,12 +68,11 @@ func writeTempFile(dir, prefix string, data []byte) (string, error) {
 		return "", err
 	}
 	_, err = file.Write(data)
-	if err1 := file.Close(); err == nil {
-		err = err1
-	}
 	if err != nil {
-		_ = os.Remove(file.Name())
-		return "", err
+		return file.Name(), err
```
peter-edge: The problem with this centralization is that this seems weird to me - on error, you're still returning me the name of the file, but what does that mean? Does the file exist? I would expect the function that creates the temp file to do the cleanup itself upon error before returning, hence the existing logic.

internal/diff/diff.go
amckinney: The `f1` type is a string, so the default value will be `""`.
It's above so that we always remove the file (even in the presence of an error. This was necessary for errors in calls to `*os.File.Close` where the file was non-nil.

amckinney:
Ah I didn't realize this was mostly copied from `gofmt`. In that case, I'll abandon this. This change wouldn't change anything besides style and breaking away from the `gofmt` code is a bad idea.

PR #84, opened 2018-05-22
Move internal/x/text to internal/text

peter-edge:
This is a helper package that handles printing, formatting, and sorting of `filename:line:column:id:message` failures.


peter-edge:
FOSSA/License Compliance should be ignored for now, per internal guidance.

internal/text/text.go
```diff
@@ -0,0 +1,284 @@
...
+		"filename": FailureFieldFilename,
+		"line":     FailureFieldLine,
+		"column":   FailureFieldColumn,
+		"id":       FailureFieldID,
+		"message":  FailureFieldMessage,
+	}
+)
+
+// FailureField references a field of a Failure.
+type FailureField int
```
amckinney: There looks like a decent amount of boilerplate around mapping these fields from their `iota` representation to their corresponding `string`. Seeing as we only ever really care about the `string` representation, should we consider simply setting these upfront?

```
// FailureField references a field of a Failure.
type FailureField string

const (
  Filename FailureField = "filename"
  Line FailureField = "line"
  Column FailureField = "column"
  ID FailureField = "id"
  Message FailureField = "message"
)
```

peter-edge: This feels pretty standard for Golang enums. Using strings allows people to get lazy and cast actual strings instead of the variable constants, which is why ints are used in Golang.

internal/text/text.go
```diff
@@ -0,0 +1,284 @@
...
+		ID:       id,
+		Filename: position.Filename,
+		Line:     position.Line,
+		Column:   position.Column,
+		Message:  fmt.Sprintf(format, args...),
+	}
+}
+
+// SortFailures sorts the Failures, by filename, line, column, id, message.
+func SortFailures(failures []*Failure) {
```
amckinney: Should we remove this function and move the `sort.Stable` call to `ParseColonSeparatedFailureFields`? 

This way none of the other packages need to explicitly call `SortFailures` and they can simply rely on it already being done. I don't see a reason where having the failures sorted would be a bad thing, so this seems useful overall.

Thoughts?

internal/text/text.go
amckinney: Works for me.

internal/text/text.go
amckinney: Ah my mistake!

PR #85, opened 2018-05-22
Move internal/x/gen to internal/cmd

peter-edge:
These just contain the helper `main` packages to generate bash completion, zsh completion, and man pages for the release tarball. `gen` wasn't the best name as there's a command `prototool gen` that has no relation to the functionality here. If there are suggestions other than `genmain`, let me know.


peter-edge:
FOSSA/License Compliance should be ignored for now, per internal guidance.

amckinney:
What are your thoughts on naming this `cmds`, or `commands`? I'd normally want to use `cmd`, but that's already taken by the Cobra wrapper. We could consider renaming the Cobra wrapper to `cli`, but I like it as `cmd`. 

It might be confusing to have both `cmd` and `{cmds, commands}`, but I think it's more clear than `genmain`.

amckinney:
We could also rename the current Cobra wrapper `cmd` package to `cobra`. Since we only ever interact with it via `cmd.Do`, `cmd.GenBashCompletion`, `cmd.GenZshCompletion`, and `cmd.GenManPages`, it might be more clear to rename these to `cobra` and rename `genmain` to `cmd`. 

It's normal for the binaries to exist under a `cmd` directory, so we should lean towards that if we aren't too tied to keeping the current `cmd` package naming.

peter-edge:
Having cmd and cmds seems really weird to me. If anything I’d just move all these gen- packages up a level to internal.

peter-edge:
I would prefer not to rename cmd for now as we already reviewed it and handed it off - id day you can make that change in the future if you want, does that seem fair?

amckinney:
Sure that works. We could also simply move each of the `gen-prototool-*` command directories under `cmd`. This would probably make sense because each of their `main` packages simply refer to `cmd` anyways. This would remove the `gen` directory entirely and consolidate it all in one place.

What are your thoughts there?

peter-edge:
Moved to `cmd`.

PR #86, opened 2018-05-23
Move internal/x/lint to internal/lint

peter-edge:


internal/lint/base_visitor.go
```diff
@@ -0,0 +1,91 @@
...
+func (baseVisitor) VisitEnum(e *proto.Enum)               {}
+func (baseVisitor) VisitComment(e *proto.Comment)         {}
+func (baseVisitor) VisitOneof(o *proto.Oneof)             {}
+func (baseVisitor) VisitOneofField(o *proto.OneOfField)   {}
+func (baseVisitor) VisitReserved(r *proto.Reserved)       {}
+func (baseVisitor) VisitRPC(r *proto.RPC)                 {}
+func (baseVisitor) VisitMapField(f *proto.MapField)       {}
+func (baseVisitor) VisitGroup(g *proto.Group)             {}
+func (baseVisitor) VisitExtensions(e *proto.Extensions)   {}
+
```
amckinney: Let's add a compile-time check for `baseVisitor` implementing the `proto.Visitor` interface:
```
var _ proto.Visitor = baseVisitor{}
```

peter-edge: Good call, I completely missed that, thanks

internal/lint/base_visitor.go
```diff
@@ -0,0 +1,91 @@
...
+type baseAddVisitor struct {
```
amckinney: Even though it directly embeds `baseVisitor`, it might be nice to include the above check for `baseAddVisitor` too.

internal/lint/base_visitor.go
```diff
@@ -0,0 +1,91 @@
...
+package lint
+
+import (
+	"text/scanner"
+
+	"github.com/emicklei/proto"
+	"github.com/uber/prototool/internal/text"
+)
+
+type baseVisitor struct{}
```
amckinney: Seeing as this is only used internally in the `newBaseAddVisitor`, it might be worth collapsing these types into one:

```
type baseVisitor struct {
  addFailure func(*text.Failure)
}
```

peter-edge: Happy to do it, I’m personally a huge fan of composition but up to you

amckinney: Yea, the composition isn't bad. @AlexAPB what are your thoughts here? I think it might be a little simpler to only have one struct, but it's a minor detail.

AlexAPB: Either work for me, but I guess this brings up a different question:
@peter-edge did you have any ideas of adding different types of linters, other than `add`s? I feel like the current behaviour is the only behaviour we'd want, no?

internal/lint/lint.go
```diff
@@ -0,0 +1,335 @@
...
+// If the config came from the settings package, this is already validated.
+func GetCheckers(config settings.LintConfig) ([]Checker, error) {
+	if len(config.IDs) == 0 && (len(config.Group) == 0 || config.Group == DefaultGroup) && len(config.IncludeIDs) == 0 && len(config.ExcludeIDs) == 0 {
+		return DefaultCheckers, nil
+	}
+
+	if len(config.IDs) > 0 {
+		var checkers []Checker
+		// n^2 woot
+		for _, checker := range AllCheckers {
```
amckinney: We could also make `AllCheckers` a map with ID keys to eliminate the iteration.
```
if checker, ok := AllCheckers[id /* config.ID */]; ok {
  checkers = append(checkers, checker)
}
```

Also, should we be returning an error if an ID is provided that we don't recognize?

internal/lint/lint.go
```diff
@@ -0,0 +1,335 @@
...
+
+// GetCheckers returns the Checkers for the LintConfig.
+//
+// The config is expected to be valid, ie slices deduped, all upper-case,
+// and only either IDs or Group/IncludeIDs/ExcludeIDs, with no overlap between
+// IncludeIDs and ExcludeIDs.
+//
+// If the config came from the settings package, this is already validated.
+func GetCheckers(config settings.LintConfig) ([]Checker, error) {
+	if len(config.IDs) == 0 && (len(config.Group) == 0 || config.Group == DefaultGroup) && len(config.IncludeIDs) == 0 && len(config.ExcludeIDs) == 0 {
```
amckinney: We might want to make a simple helper function for these types of conditionals:
```
if config.IsEmpty() {
  return DefaultCheckers, nil
}
```

internal/lint/lint.go
```diff
@@ -0,0 +1,335 @@
...
+
+	checkersMap := make(map[string]Checker, len(baseCheckers))
+	for _, checker := range baseCheckers {
+		checkersMap[checker.ID()] = checker
+	}
+	for _, excludeID := range config.ExcludeIDs {
+		delete(checkersMap, excludeID)
+	}
+	// n^2 woot
+	for _, checker := range AllCheckers {
```
amckinney: Similar to above comment.

internal/lint/lint.go
amckinney: `IsEmpty` might not be the best name here.

internal/lint/base_visitor.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (v baseAddVisitor) AddFailuref(position scanner.Position, format string, args ...interface{}) {
+	v.add(text.NewFailuref(position, "", format, args...))
+}
+
+// extendedVisitor extends the proto.Visitor interface.
+// extendedVisitors are expected to be called with one file at a time,
+// and are not thread-safe.
+type extendedVisitor interface {
```
AlexAPB: micro nit: maybe something like `visitorLifecycle`? Just seems like odd naming.

amckinney:
Package looks good overall! Some minor comments.

internal/lint/base_visitor.go
peter-edge: Yea I did, I didn't end up doing so. We can work to just make it add, but up to you - since the abstraction is already there, might be nice to leave it, but also might be nice to simplify.

internal/lint/base_visitor.go
```diff
@@ -0,0 +1,91 @@
...
+func (baseVisitor) VisitNormalField(i *proto.NormalField) {}
+func (baseVisitor) VisitEnumField(i *proto.EnumField)     {}
+func (baseVisitor) VisitEnum(e *proto.Enum)               {}
+func (baseVisitor) VisitComment(e *proto.Comment)         {}
+func (baseVisitor) VisitOneof(o *proto.Oneof)             {}
+func (baseVisitor) VisitOneofField(o *proto.OneOfField)   {}
+func (baseVisitor) VisitReserved(r *proto.Reserved)       {}
+func (baseVisitor) VisitRPC(r *proto.RPC)                 {}
+func (baseVisitor) VisitMapField(f *proto.MapField)       {}
+func (baseVisitor) VisitGroup(g *proto.Group)             {}
```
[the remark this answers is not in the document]
peter-edge: We lint for proto3, but we support proto2 in Prototool.

internal/lint/check_message_names_camel_case.go
```diff
@@ -0,0 +1,54 @@
...
+type messageNamesCamelCaseVisitor struct {
+	baseAddVisitor
+}
+
+func (v messageNamesCamelCaseVisitor) VisitMessage(message *proto.Message) {
+	// for nested messages
+	for _, child := range message.Elements {
+		child.Accept(v)
+	}
+	if message.IsExtend {
```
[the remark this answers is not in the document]
peter-edge: We lint for proto3, but support proto2 in Prototool.

internal/lint/base_visitor.go
peter-edge: Done

internal/lint/lint.go
peter-edge: I think I did that originally, but it ended up being weirder to read than this.

internal/lint/lint.go
peter-edge: Same answer, see above

internal/lint/base_checker.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"strings"
+	"sync"
+
+	"github.com/emicklei/proto"
+	"github.com/uber/prototool/internal/text"
+)
+
+type baseChecker struct {
```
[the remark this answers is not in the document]
peter-edge: I thought `golint` would complain about that, but I was wrong (I just tested it). It's going to be a gigantic pain to change `Checker` to `Linter` and `Check` to `Lint`, but if both you and @amckinney want me to do it, might as well do it now (will wait on @amckinney).

internal/lint/base_checker.go
```diff
@@ -0,0 +1,90 @@
...
+	id      string
```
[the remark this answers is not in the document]
peter-edge: I want to keep it as ID, as this matches with the `internal/text` package. I could do `lintID` perhaps, @amckinney thoughts?

internal/lint/lint.go
```diff
@@ -0,0 +1,335 @@
...
+	Check(dirPath string, descriptors []*proto.Proto) ([]*text.Failure, error)
+}
+
+// NewChecker is a convenience function that returns a new Checker for the
+// given parameters.
+//
+// The ID will be upper-cased.
+//
+// Failures returned from check do not need to set the ID, this will be overwritten.
+func NewChecker(id string, purpose string, check func(string, []*proto.Proto) ([]*text.Failure, error)) Checker {
```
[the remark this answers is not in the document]
peter-edge: Yea see above

internal/lint/base_checker.go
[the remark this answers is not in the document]
peter-edge:
```
$ ag checker | wc -l
     184
```

internal/lint/base_visitor.go
peter-edge: Idk, up to you. It's not a lifecycle, its a visitor with extra stuff.

internal/lint/base_visitor.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/emicklei/proto"
+	"github.com/uber/prototool/internal/text"
+)
+
+type baseVisitor struct{}
+
+func (baseVisitor) OnStart(*proto.Proto) error { return nil }
+func (baseVisitor) Finally() error             { return nil }
+
+func (baseVisitor) VisitMessage(m *proto.Message)         {}
```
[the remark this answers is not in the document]
peter-edge: This is the order that the functions are in the interface definition I believe https://godoc.org/github.com/emicklei/proto#Visitor

internal/lint/check_comments_no_c_style.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+func (v commentsNoCStyleVisitor) VisitReserved(element *proto.Reserved) {
+	v.checkComments(element.Position, element.Comment, element.InlineComment)
+}
+
+func (v commentsNoCStyleVisitor) VisitRPC(element *proto.RPC) {
+	v.checkComments(element.Position, element.Comment, element.InlineComment)
+	for _, child := range element.Options {
```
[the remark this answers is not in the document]
peter-edge: Yea I was the one that deprecated it basically. This still works, I can update if you want, I just didn't want to touch this.

internal/lint/check_comments_no_c_style.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (v commentsNoCStyleVisitor) VisitMapField(element *proto.MapField) {
+	v.checkComments(element.Position, element.Comment, element.InlineComment)
+	for _, child := range element.Options {
+		child.Accept(v)
+	}
+}
+
+func (v commentsNoCStyleVisitor) VisitGroup(element *proto.Group) {
```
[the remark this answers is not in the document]
peter-edge: See above

internal/lint/check_file_options_same_in_dir.go
```diff
@@ -0,0 +1,107 @@
...
+			fileOption:       fileOption,
+			fileOptionValues: make(map[string]struct{}),
+		}
+		if err := runVisitor(visitor, descriptors); err != nil {
+			return err
+		}
+		if len(visitor.fileOptionValues) > 1 {
+			if _, ok := visitor.fileOptionValues[""]; ok {
+				for _, descriptor := range descriptors {
+					visitor.AddFailuref(scanner.Position{Filename: descriptor.Filename}, "File option %q set in some files in directory but not in others.", fileOption)
```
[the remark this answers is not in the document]
peter-edge: Really? lol :-p

internal/lint/check_request_response_types_unique.go
```diff
@@ -0,0 +1,64 @@
...
+package lint
+
+import (
+	"github.com/emicklei/proto"
+	"github.com/uber/prototool/internal/text"
+)
+
+var requestResponseTypesUniqueChecker = NewAddChecker(
+	"REQUEST_RESPONSE_TYPES_UNIQUE",
+	"Verifies that all request and response types are unique to each rpc.",
```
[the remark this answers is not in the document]
peter-edge: Done

internal/lint/check_comments_no_c_style.go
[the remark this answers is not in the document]
peter-edge: Done

peter-edge:
All PR comments are addressed, there's a bunch of responses to the comments, let me know!

internal/lint/lint.go
amckinney: Not sure if I follow here. Would we just need to range over the `config.IDs` and check to see if the key exists?

```
var checkers []Checker
for _, id := range config.IDs {
  if checker, ok := AllCheckers[id]; ok {
    checkers = append(checkers, checker)
  }
}
```

internal/lint/base_checker.go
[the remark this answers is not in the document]
AlexAPB: sounds good to me!

internal/lint/check_file_options_same_in_dir.go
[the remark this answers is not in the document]
AlexAPB: haha just a nit since the amount of nesting pushes it outwards. No worries if you think it's fine as is.

internal/lint/check_message_names_camel_case.go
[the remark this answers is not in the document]
AlexAPB: yeah, i'm of the same mindset as @amckinney, i'd rather keep this tool as 'forward looking' as possible, even if it is fairly trivial to support some proto2 functionality

internal/lint/lint.go
AlexAPB: I see the confusion here.
@amckinney, `AllCheckers` is initialized [here](https://sourcegraph.com/github.com/uber/prototool@d8678e3fe743bede4a7ee98bd9d15f396517fc81/-/blob/internal/lint/lint.go#L37:2) so we can't create that mapping in the same way that we're creating that slice.
@peter-edge what if we create a new map (eg `allCheckers[id]Checker`) before these calls, then we can reduce the n^2 to n

internal/lint/base_visitor.go
AlexAPB: I'm of the mindset for simplifying if possible

internal/lint/base_checker.go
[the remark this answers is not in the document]
peter-edge: Yea it'll be fine. I'll do it!

internal/lint/check_message_names_camel_case.go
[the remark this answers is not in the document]
peter-edge: So I think there's a bit of misunderstanding here - proto2 isn't something just of the past, we basically have to support the syntax. Even if we only use protobuf 3.0.0 and above, proto2 files are all over the place. For example, the Well-Known Types https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto are proto2. Instead of thinking it as a breaking version of the language, it's more like "we only use go 1.10 and above to build prototool, but you can use elements of go 1.4" (bad example but).

internal/lint/base_checker.go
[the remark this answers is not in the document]
peter-edge: Done

internal/lint/lint.go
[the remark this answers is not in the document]
peter-edge: Done

internal/lint/lint.go
peter-edge: We could. Just want to point out that this takes on the order of microseconds (I haven't benchmarked it, but I can almost guarantee that lol), the number of checkers is not going to explode to be 1,000,000 checkers. I rather keep it simple personally, but I can change this if you want.

peter-edge:
All comments should be addressed again! @AlexAPB @amckinney

internal/lint/check_message_names_camel_case.go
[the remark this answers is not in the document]
AlexAPB: gotcha

internal/lint/lint.go
AlexAPB: Yeah true, i'm indifferent. @amckinney your call, since you brought it up

amckinney: Got it, yea this is fine as is. :+1:

internal/lint/check_message_names_camel_case.go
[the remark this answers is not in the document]
amckinney: Ah ok makes sense. Thanks for the clarification.

PR #87, opened 2018-05-23
Assign IDs to all Failures; refactor internal/text and move to internal/failure

amckinney:
Note that this moves the `internal/text` package to `internal/failure`. Seeing as this package is primarily concerned with failures, it makes more sense to remove some of the verbose naming (e.g. `FailureField`, `FailureWriter`, etc.).

The primary change introduced here is the `failure.ID` type, which is used to categorize prototool failures. The available IDs are `{Error, Format, Lint, Proto}`. Previously, we supported `DIFF_FORMAT` and `INVALID_PROTOBUF` for special cases. These are now `FORMAT` and `PROTO`, respectively. All of the failures created in `internal/x/lint` are now identified by `LINT` and those created in `internal/x/protoc` are `PROTO`.

The mapping is now as follows:

```
Error -> "ERROR"
Format -> "FORMAT"
Lint -> "LINT"
Proto -> "PROTO"
```

This now enforces that all failures creates with `failure.Newf` will contain a valid `failure.ID`. And in the cases where `failure.Newf` cannot be easily used (in the absence of a `scanner.Position`), such as in `internal/x/protoc`, we use a `newFailure` wrapper/helper.

We continue to use the fallback behavior previously found in `failure.Fprintln` so that we make sure all of our failures will contain valid fields. This is especially important if any `failure.Failure` types are created without a filename, line, or column (i.e. without using `failure.Newf`). 

With that said, we should avoid creating `failure.Failure` types in-place whenever possible. Rather, we should rely on the `failure.Newf` function so that we always categorize our failures with a `failure.ID`. Ideally, the fallback `failure.Error` ID would not be used.

-------------

Note that I've attempted to break this up in more-easily reviewable commits so that the changes here are better understood.

### DIFF
--- a/internal/cmd/cmd_test.go
+++ b/internal/cmd/cmd_test.go
@@ -54,39 +54,39 @@ func TestCompile(t *testing.T) {
 	assertDoCompileFiles(
 		t,
 		false,
-		`testdata/compile/dep_errors.proto:6:1:Expected ";".`,
+		`testdata/compile/dep_errors.proto:6:1:PROTO Expected ";".`,
 		"testdata/compile/dep_errors.proto",
 	)
 	assertDoCompileFiles(
 		t,
 		false,
-		`dep_errors.proto:6:1:Expected ";".
-		testdata/compile/errors_on_import.proto:10:3:"foo.DepError" is not defined.`,
+		`dep_errors.proto:6:1:PROTO Expected ";".
+		testdata/compile/errors_on_import.proto:10:3:PROTO "foo.DepError" is not defined.`,
 		"testdata/compile/errors_on_import.proto",
 	)
 	assertDoCompileFiles(
 		t,
 		false,
-		`testdata/compile/extra_import.proto:1:1:Import "dep.proto" was not used.`,
+		`testdata/compile/extra_import.proto:1:1:PROTO Import "dep.proto" was not used.`,
 		"testdata/compile/extra_import.proto",
 	)
 	assertDoCompileFiles(
 		t,
 		false,
-		`testdata/compile/json_camel_case_conflict.proto:1:1:The JSON camel-case name of field "helloworld" conflicts with field "helloWorld". This is not allowed in proto3.`,
+		`testdata/compile/json_camel_case_conflict.proto:1:1:PROTO The JSON camel-case name of field "helloworld" conflicts with field "helloWorld". This is not allowed in proto3.`,
 		"testdata/compile/json_camel_case_conflict.proto",
 	)
 	assertDoCompileFiles(
 		t,
 		false,
-		`testdata/compile/missing_package_semicolon.proto:5:1:Expected ";".`,
+		`testdata/compile/missing_package_semicolon.proto:5:1:PROTO Expected ";".`,
 		"testdata/compile/missing_package_semicolon.proto",
 	)
 	assertDoCompileFiles(
 		t,
 		false,
-		`testdata/compile/missing_syntax.proto:1:1:No syntax specified. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version.
-		testdata/compile/missing_syntax.proto:4:3:Expected "required", "optional", or "repeated".`,
+		`testdata/compile/missing_syntax.proto:1:1:PROTO No syntax specified. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version.
+		testdata/compile/missing_syntax.proto:4:3:PROTO Expected "required", "optional", or "repeated".`,
 		"testdata/compile/missing_syntax.proto",
 	)
 	assertDoCompileFiles(
@@ -98,7 +98,7 @@ func TestCompile(t *testing.T) {
 	assertDoCompileFiles(
 		t,
 		false,
-		`testdata/compile/not_imported.proto:11:3:"foo.Dep" seems to be defined in "dep.proto", which is not imported by "not_imported.proto".  To use it here, please add the necessary import.`,
+		`testdata/compile/not_imported.proto:11:3:PROTO "foo.Dep" seems to be defined in "dep.proto", which is not imported by "not_imported.proto".  To use it here, please add the necessary import.`,
 		"testdata/compile/dep.proto",
 		"testdata/compile/not_imported.proto",
 	)
--- a/internal/failure/failure.go
+++ b/internal/failure/failure.go
@@ -113,13 +113,13 @@ type Writer interface {
 }
 
 // Fprintln prints the Failure to the writer with the given ordered fields.
+// The given fields will overwrite the fields already set in this Failure.
 func (f *Failure) Fprintln(writer Writer, fields ...Field) error {
 	if len(fields) == 0 {
 		fields = _defaultFields
 	}
-	written := false
+	printColon := true
 	for i, field := range fields {
-		printColon := true
 		switch field {
 		case Filename:
 			filename := f.Filename
@@ -129,7 +129,6 @@ func (f *Failure) Fprintln(writer Writer, fields ...Field) error {
 			if _, err := writer.WriteString(filename); err != nil {
 				return err
 			}
-			written = true
 		case Line:
 			line := strconv.Itoa(f.Line)
 			if line == "0" {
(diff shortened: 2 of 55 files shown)


internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+		if err != nil {
+			return nil, err
+		}
+		failureFields[i] = ff
+	}
+	return failureFields, nil
+}
+
+// Failure is a failure with a position in text.
+type Failure struct {
```
peter-edge: This will fail lint as Golang says foo.Foo is not allowed. For this and other reasons ill elaborate on tomorrow, I think this should be reverted to text. Failure is a noun, which doesn’t work well for packages.

internal/cmd/cmd_test.go
```diff
@@ -54,39 +54,39 @@ func TestCompile(t *testing.T) {
 	assertDoCompileFiles(
 		t,
 		false,
-		`testdata/compile/dep_errors.proto:6:1:Expected ";".`,
+		`testdata/compile/dep_errors.proto:6:1:PROTO Expected ";".`,
```
peter-edge: You do not want to print this out. This is a `protoc` error, and the goal is to just convert them to `filename:line:column:message`. The logic that prints this out should be removed.

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package failure
```
peter-edge: Nouns aren't great package names, especially as we have a struct named `Failure`. Having `failure.Failure` creates weirdness, such as:

```
failure := &failure.Failure{} // if you name the variable failure, which we do a lot in prototool, or failures
failure.Sort(failures) // if you have a slices of failures named failures
```

I would keep this as `text`, or another name.

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+	// Field references a Failure field.
+	Field int
+
+	// ID references a Failure identifier.
+	ID int
+)
+
+const (
+	// Filename references the Filename field of a Failure.
+	Filename Field = iota
```
peter-edge: I generally do verbose naming on enums on purpose - this introduces other weirdness:

```
var failure failure.Failure // a struct of type failure.Failure
_ = failure.Filename // is this the enum Field value Filename, or the struct field failure.Filename
```

What is the benefit of removing the `FailureField` prefix?

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"text/scanner"
+)
+
+type (
+	// Field references a Failure field.
+	Field int
```
peter-edge: This is a field of a `Failure`, which is a named struct, therefore its a `FailureField`, not a `Field` IMO.

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+				return err
+			}
+			if _, err := writer.WriteString(f.Message); err != nil {
+				return err
+			}
+			printColon = false
+		default:
+			return fmt.Errorf("unknown Field: %v", field)
+		}
+		if printColon && i != len(fields)-1 {
```
peter-edge: By removing the `written` boolean, you now print a colon even if the field was empty. The `written` boolean should be re-introduced.

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+		Message,
+	}
+	_stringToField = map[string]Field{
+		"filename": Filename,
+		"line":     Line,
+		"column":   Column,
+		"id":       Identifier,
+		"message":  Message,
+	}
+	_IDToString = map[ID]string{
```
peter-edge: What is this doing specifically? The ID is primarily meant to reference lint IDs, which these are not. If you were to enumerate all possible IDs, this would be in the `lint` package, which should not be a dependency of this package.

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+				return err
+			}
+		}
+	}
+	_, err := writer.WriteRune('\n')
+	return err
+}
+
+// Newf is a helper that returns a new Failure.
+func Newf(position scanner.Position, id ID, format string, args ...interface{}) *Failure {
```
peter-edge: We have multiple types in this package, I personally do not like the convention of `New` as a package-level method. A new of what? That's how I read it when I look at this being used in other parts of the code.

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+		Line:     position.Line,
+		Column:   position.Column,
+		Message:  fmt.Sprintf(format, args...),
+	}
+}
+
+// Sort sorts the Failures by the following precedence:
+//
+//  filename > line > column > id > message
+func Sort(fs []*Failure) {
```
peter-edge: Same as `Newf`. Also see the comment about `failure.Sort(failures)`.

internal/failure/failure.go
```diff
@@ -0,0 +1,268 @@
...
+	"strings"
+	"text/scanner"
+)
+
+type (
+	// Field references a Failure field.
+	Field int
+
+	// ID references a Failure identifier.
+	ID int
```
peter-edge: We should discuss offline what an ID is, this isn't actually encapsulating that concept. An ID is meant to reference a linter primarily, and the below does not enumerate the values. IDs are generally freeform strings.

internal/failure/failure.go
peter-edge: Look at `internal/x/lint/base_checker.go`, function `baseChecker#Check`. The issue here is that the concept of ID is getting overloaded. We should either have a lint ID, or not. If anything, we should remove `INVALID_PROTOBUF`, etc.

peter-edge:
Going over this again, I think this is the issue:

> All of the failures created in internal/x/lint are now identified by LINT and those created in internal/x/protoc are PROTO.

This is not true. See `baseChecker` in the lint package:

```go
func (c *baseChecker) Check(dirPath string, descriptors []*proto.Proto) ([]*text.Failure, error) {
	failures, err := c.check(dirPath, descriptors)
	for _, failure := range failures {
		failure.ID = c.id
	}
	return failures, err
}
```

The purpose of the ID field is to reference lint failure types for automated identification. There are many ID types. This PR changes that, and would change the meaning of ID, which breaks a lot of functionality. You really either want an ID (if you have a lint failure that can be categorized) or don't (in the case of protoc failures).

internal/cmd/cmd_test.go
peter-edge: I see why this is printed out actually - we'll discuss. I think you're right that everything should have an ID, but see discussion below of what an ID is, we'll discuss offline.

internal/failure/failure.go
peter-edge: As an example, here's what `prototool lint` prints out now:

```
$ prototool lint internal/cmd/testdata/lint/lots.proto
internal/cmd/testdata/lint/lots.proto:1:1: File option "go_package" is required.
internal/cmd/testdata/lint/lots.proto:1:1: File option "java_package" is required.
```

Note the extra space after `1:1:`. In the old version, this was:

```
$ prototool lint internal/cmd/testdata/lint/lots.proto
internal/cmd/testdata/lint/lots.proto:1:1:File option "go_package" is required.
internal/cmd/testdata/lint/lots.proto:1:1:File option "java_package" is required.
```

amckinney:
Discussed offline - 

This PR brought to light a variety of potential areas of concern, especially with regard to `text.Failure` creation and standardizing the `text.Failure.ID` attribute. I'll open an issue that details some of the considerations offered here so that we can address them at a later time (if necessary).

PR #88, opened 2018-05-24
Add linters to validate java_multiple_files and java_outer_classname are unset

peter-edge:
This adds linters to validate that `java_multiple_files` and `java_outer_classname` are unset. We used to recommend setting these but do not anymore, so for consistency, we want to make sure they are unset. This also removes these declarations from lint tests where they were added for when we did require them, and adds testing to make sure the proper errors are produced.


internal/lint/check_file_options_unset.go
```diff
@@ -0,0 +1,79 @@
...
+
+	fileOption string
+
+	seen     bool
+	position scanner.Position
+}
+
+func (v *fileOptionsUnsetVisitor) OnStart(descriptor *proto.Proto) error {
+	v.seen = false
+	v.position = scanner.Position{}
```
amckinney: I don't know if these are actually necessary. `false` and `scanner.Position` should both be the default values for `*fileOptionsUnsetVisitor`. I think we can omit the `OnStart` method entirely (unless it's required to implement some interface. In that case, it could be:

```
func (v *fileOptionsUnsetVisitor) OnStart(descriptor *proto.Proto) error { return nil }
```

internal/lint/check_file_options_unset.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"text/scanner"
+
+	"github.com/emicklei/proto"
+	"github.com/uber/prototool/internal/text"
+)
+
+var fileOptionsUnsetJavaMultipleFilesChecker = NewAddChecker(
+	"FILE_OPTIONS_UNSET_JAVA_MULTIPLE_FILES",
```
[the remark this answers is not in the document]
amckinney: Got it, thanks for the example here.

internal/lint/check_file_options_unset.go
peter-edge: Right now we re-use these linters, which admittedly isn’t a great pattern, so they need to be reset each time.

amckinney: Ok, good to know 👍

AlexAPB: This is unfortunate that linters are reused, but I can understand that it may not make much sense to rebuild them for every file -- @peter-edge you didn't by chance do any benchmarking on this did you :D

peter-edge: No, but the time this all takes is very trivial, like on the order of milliseconds

PR #94, opened 2018-06-01
Move internal/x/grpc to internal/grpc

amckinney:


peter-edge:
I'm going to close this for now so that we can make changes like #98 without having massive conflicts. This will probably be the last package we review as it is.

Issue #95, opened 2018-06-01
grpc - temporary protobuf descriptors are not cleaned up


twilly:
`prototool grpc` leaves behind temporary protobuf descriptors on every call.

To reproduce, make any call with `grpc` and inspect your OS's typical temp file location. For example:

```
$ prototool grpc foo.Bar/Baz ''
$ ls /tmp/prototool*
/tmp/prototool661509915  /tmp/prototool892888492
```

peter-edge:
Closed by #96

peter-edge:
This is fixed on v0.3.0.

PR #96, opened 2018-06-01
protoc.Compile: clean up cmdMetas

twilly:
Deferring the call to `cleanCmdMetas` so early means we'll be calling
with a nil value for `allCmdMetas` rather than the final pointer value
that `append` could have returned.


twilly:
This should fix Issue #95.

PR #102, opened 2018-06-11
Move internal/x/format to internal/format

peter-edge:


internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	// https://github.com/emicklei/proto/commit/5a91db7561a4dedab311f36304fcf0512343a9b1
+	//if inlineComment.Cstyle {
+	//args = append(args, inlineComment.Lines[0])
+	//v.P(args...)
+	//for _, line := range inlineComment.Lines[1:] {
+	//v.P(line)
+	//}
+	//return
+	//}
```
amckinney: Can we remove these commented out lines?

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+//func (v *baseVisitor) VisitEnumField(element *proto.EnumField)     {}
+//func (v *baseVisitor) VisitEnum(element *proto.Enum)               {}
+//func (v *baseVisitor) VisitComment(element *proto.Comment)         {}
+//func (v *baseVisitor) VisitOneof(element *proto.Oneof)             {}
+//func (v *baseVisitor) VisitOneofField(element *proto.OneOfField)   {}
+//func (v *baseVisitor) VisitReserved(element *proto.Reserved)       {}
+//func (v *baseVisitor) VisitRPC(element *proto.RPC)                 {}
+//func (v *baseVisitor) VisitMapField(element *proto.MapField)       {}
+//func (v *baseVisitor) VisitGroup(element *proto.Group)             {}
+//func (v *baseVisitor) VisitExtensions(element *proto.Extensions)   {}
```
amckinney: Do we need to implement the `emicklei/proto.Visitor` interface here, or can we remove these?

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+	args = append(args, ` //`, cleanCommentLine(inlineComment.Lines[0]))
+	v.P(args...)
+	for i, line := range inlineComment.Lines[1:] {
+		line = cleanCommentLine(line)
+		if line == "" && i != len(inlineComment.Lines)-1 {
+			v.P(`//`)
+		} else {
+			v.P(`//`, line)
+		}
+	}
```
amckinney: This might be equivalent to:
```
for i, line := range inlineComment.Lines[1:] {
  v.P(`//`, cleanCommentLine(line))
}
```

Your first conditional will only ever run with `v.P("//")` which should be equivalent to `v.P("//", "")` or just `v.P("//", line)` in that case.

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+	// https://github.com/emicklei/proto/commit/5a91db7561a4dedab311f36304fcf0512343a9b1
+	//if comment.Cstyle {
+	//for _, line := range comment.Lines {
+	//v.P(line)
+	//}
+	//return
+	//}
+	// this is weird for now
+	// we always want non-c-style after formatting
```
amckinney: remove

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+	for i, line := range comment.Lines {
+		line = cleanCommentLine(line)
+		if line == "" && !(i == 0 || i == len(comment.Lines)-1) {
+			v.P(`//`)
+		} else {
+			v.P(`//`, line)
+		}
```
amckinney: Similar to above - we might be able to shorten this.

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+			// if empty, we do not want to print the key or empty value
+			if source := o.Constant.SourceRepresentation(); source != "" {
+				v.PWithInlineComment(o.InlineComment, prefix, o.Name, ` = `, source, suffix)
+			}
+		} else if len(o.Constant.Array) > 0 { // both Array and OrderedMap should not be set simultaneously, need more followup with emicklei/proto
+			v.Failures = append(
+				v.Failures,
+				text.NewFailuref(o.Position, "INVALID_PROTOBUF", "top-level options should never be arrays, this should not compile with protoc"),
+			)
+		} else { // len(o.Constant.OrderedMap) > 0
```
amckinney: Do we need this conditional comment? Was this necessary before?

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+			for _, namedLiteral := range o.Constant.OrderedMap {
+				v.pInnerLiteral(namedLiteral.Name, *namedLiteral.Literal, "")
+			}
+			v.Out()
+			v.PWithInlineComment(o.InlineComment, `}`, suffix)
+		}
+	}
+}
+
+// should only be called by Poptions
```
amckinney: nit: `POptions`

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+		for i, iLiteral := range literal.Array {
+			iSuffix := ""
+			if len(literal.Array) > 1 && i != len(literal.Array)-1 {
+				iSuffix = ","
+			}
+			v.pInnerLiteral("", *iLiteral, iSuffix)
+		}
+		v.Out()
+		v.P(`]`, suffix)
+	} else { // len(literal.OrderedMap) > 0
```
amckinney: Similar to above comment.

internal/format/base_visitor.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	v.P(prefix, t, " ", field.Name, " = ", field.Sequence, " [")
+	v.In()
+	v.POptions(true, field.Options...)
+	v.Out()
+	v.PWithInlineComment(field.InlineComment, "];")
+}
+
+func cleanCommentLine(line string) string {
+	// TODO: this is not great
```
amckinney: Meh, I see what you're doing here. Seems fine for now.

internal/format/first_pass_visitor.go
```diff
@@ -0,0 +1,187 @@
...
+func (v *firstPassVisitor) VisitEnumField(element *proto.EnumField) {
+	v.haveHitNonComment = true
+}
+
+func (v *firstPassVisitor) VisitEnum(element *proto.Enum) {
+	v.haveHitNonComment = true
+}
+
+func (v *firstPassVisitor) VisitComment(element *proto.Comment) {
+	if !v.haveHitNonComment {
```
amckinney: Ah, so does this mean the first pass visitor does not concern itself with all of the comments in the file (unless it's the first thing it hits)? And then the middleVisitor picks it up from there? 

Could we leave some descriptive comments in the type declaration for what these types concern themselves with when passing over and formatting the protobuf file?

internal/format/first_pass_visitor.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+		v.PWithInlineComment(i.InlineComment, `import `, kind, `"`, i.Filename, `";`)
+	}
+}
+
+func isProbablyCustomOption(option *proto.Option) bool {
+	// you can technically do ie google.protobuf.java_package
+	// but we're not going to handle this as I mean come on
+	return strings.HasPrefix(option.Name, "(")
+}
```
amckinney: Strange naming here - should we just call this `isCustomOption`? I don't like the "maybe" feeling :P

internal/format/middle_visitor.go
```diff
@@ -0,0 +1,306 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/emicklei/proto"
+	"github.com/uber/prototool/internal/settings"
+	"github.com/uber/prototool/internal/text"
+)
+
+type middleVisitor struct {
```
amckinney: Seeing as there are only two visitors, what are your thoughts on renaming these types to `preVisitor` and `postVisitor`, or something similar? We could leave it as `firstVisitor` and rename this to `lastVisitor`?

Or, depending on if we add context to these types (mentioned above), maybe a more descriptive name for these types will come out of it.

internal/format/middle_visitor.go
```diff
@@ -0,0 +1,306 @@
...
+		v.PComment(element)
+		v.P()
+	}
+}
+
+func (v *middleVisitor) VisitOneof(element *proto.Oneof) {
+	v.haveHitNonComment = true
+	v.PComment(element.Comment)
+	if len(element.Elements) == 0 {
+		// TODO: is this even legal?
```
amckinney: We should figure this out and see if it's necessary after all.

internal/format/middle_visitor.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (v *middleVisitor) VisitNormalField(element *proto.NormalField) {
+	v.haveHitNonComment = true
+	prefix := ""
+	if element.Repeated {
+		prefix = "repeated "
+	}
+	if v.isProto2 {
+		// technically these are only set if the file is proto2
+		// but doing this just to make srue
```
amckinney: nit: srue -> sure

amckinney:
Left some minor comments. Looking good overall.

internal/format/first_pass_visitor.go
AlexAPB: yeah this would be super helpful, it wasn't clear why we needed these initially

internal/format/base_visitor.go
peter-edge: Done

internal/format/base_visitor.go
peter-edge: Done

internal/format/base_visitor.go
peter-edge: I think it's actually just `range inlineComment.Lines`, not cut off, right? Am I missing something? Not sure what I was doing here.

So:

```go
for i, line := range inlineComment.Lines {
  v.P(`//`, cleanCommentLine(line))
}
```

internal/format/base_visitor.go
peter-edge: Done

internal/format/base_visitor.go
peter-edge: See above, want to make sure we're right on this.

internal/format/base_visitor.go
peter-edge: Yes, you need to figure out if one of the two as length > 0. This is what this `else` is doing, so want to explicitly comment it.

internal/format/base_visitor.go
peter-edge: Done

internal/format/base_visitor.go
peter-edge: Same answer

internal/format/first_pass_visitor.go
```diff
@@ -0,0 +1,187 @@
...
+func newFirstPassVisitor(config settings.Config) *firstPassVisitor {
+	return &firstPassVisitor{baseVisitor: newBaseVisitor(config.Format.Indent)}
+}
+
+func (v *firstPassVisitor) Do() []*text.Failure {
+	if v.Syntax != nil {
+		v.PComment(v.Syntax.Comment)
+		if v.Syntax.Comment != nil {
+			// special case
+			v.P()
```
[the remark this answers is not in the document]
peter-edge: Usually the `syntax` line is the first line in the file. In the case there is a comment at the top of the file, I always want a newline between `syntax` and that comment. So:

```
// bla bla
syntax = "proto3";
```

Converts to:
```
// bla bla

syntax = "proto3";
```

The only time you'd generally comment above syntax is for a LICENSE header or a file-level comment. We could make it so that this doesn't format into a new line, but this was the thinking.

internal/format/first_pass_visitor.go
peter-edge: This means the only file-level comments that are printed before syntax, file-level options, imports, package are ones at the top of the file. Commenting.

internal/format/first_pass_visitor.go
peter-edge: I'm just going to get rid of the distinction and let sort take care of it.

internal/format/middle_visitor.go
peter-edge: Yea I was going to have a visitor at the end, just never renamed. How about `firstPassVisitor` and `mainVisitor`?

internal/format/middle_visitor.go
peter-edge: Yea we should, lol.

internal/format/middle_visitor.go
peter-edge: Idk how you find all these comment spelling errors, it's impressive

internal/format/format.go
```diff
@@ -0,0 +1,56 @@
...
+
+package format
+
+import (
+	"github.com/uber/prototool/internal/settings"
+	"github.com/uber/prototool/internal/text"
+	"go.uber.org/zap"
+)
+
+// TODO: single primitive options on same line
```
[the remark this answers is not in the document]
peter-edge: This is so if you have a field:

```
int woot = 5 [ (bar.field_option) = true ];
```

That is printed out instead of the current:

```
int woot = 5 [
  (bar.field_option) = true
];
```

Filed #108 instead of this TODO.

internal/format/format.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Transformer transforms an input file into an output file.
+type Transformer interface {
+	// Transform transforms the data.
+
```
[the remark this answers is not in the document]
peter-edge: Fixed

internal/format/transformer.go
```diff
@@ -0,0 +1,100 @@
...
+	for _, element := range descriptor.Elements {
+		element.Accept(logVisitor)
+	}
+
+	firstPassVisitor := newFirstPassVisitor(config)
+	for _, element := range descriptor.Elements {
+		element.Accept(firstPassVisitor)
+	}
+	failures := firstPassVisitor.Do()
+	buffer := bytes.NewBuffer(nil)
```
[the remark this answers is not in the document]
peter-edge: While this will technically work here, the returned `Bytes()` value is only valid until the next modification in `firstPassVisitor` - while we don't do another modification, I don't like relying on these semantics and prefer to make a copy here.

internal/format/printer.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// In adds one indent.
+func (p *printer) In() {
+	p.indentCount++
+}
+
+// Out deletes one indent.
+func (p *printer) Out() {
+	// might want to error
```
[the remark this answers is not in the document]
peter-edge: I mean maybe, but then I have to weave it all through to this simple helper class. If this happens, you'll notice in your outputted files, this isn't minor.

peter-edge:
All comments addressed @AlexAPB @amckinney

internal/format/middle_visitor.go
AlexAPB: sgtm

internal/format/middle_visitor.go
AlexAPB: Just checked with the `go`, `gogo`, `python`, `java` protoc plugin, this **is not** allowed: 
```
foo.proto:12:19: Expected type name.
```

Although, it may be worth it to keep this line as that behavior seems reasonable / not surprising, if other plugins allow this?

internal/format/printer.go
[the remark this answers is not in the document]
AlexAPB: yeah true

internal/format/transformer.go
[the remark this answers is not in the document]
AlexAPB: yeah good call, just read the `NewBuffer` doc

internal/format/base_visitor.go
peter-edge: Argh jeez, had spots mixed up cuz commits all over the place, fixing.

internal/format/base_visitor.go
peter-edge: Done

internal/format/middle_visitor.go
peter-edge: Done

internal/format/first_pass_visitor.go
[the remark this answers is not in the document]
peter-edge: Done

peter-edge:
All comments addressed, will merge if CI passes.

PR #104, opened 2018-06-13
gRPC JSON Output Flag

peter-edge:
Follow-up to #98.

This adds a flag `--json-output` that will output all results in a JSON message with keys `headers, trailers, response, status`. This means if you want to just get the response, for example, you can pipe the result to jq by doing `| jq .response`.

The hackiness is because we have to use a different JSON Marshaller for the Protobuf messages than for the top-level wrapper message. I'm sure there's cleaner ways to do this.


peter-edge:
Closing for now, filed an issue to track this.

peter-edge:
Re-opening this because I think it's extremely useful. The messiness is almost unavoidable now that I think of it again - Golang doesn't let you "sub-marshal" fields to the best of my knowledge, so if you want a field that is already marshaled to JSON to not be escaped out, you need to do some stuff (maybe I'm wrong though, it would be great to be corrected on this).

I would recommend testing this using the example.

```
# both terminals have the prototool repo as the current directory

# in another terminal
$ go run example/cmd/excited/main.go

# in your current terminal
$ prototool grpc example --address 0.0.0.0:8080 --method foo.ExcitedService/Exclamation --data '{"value":"hello"}'
{
  "value": "hello!"
}
$ prototool grpc example --address 0.0.0.0:8080 --method foo.ExcitedService/Exclamation --data '{"value":"hello"}' --json-output
{"headers":{"content-type":["application/grpc"]}}
{"response":{"value":"hello!"}}
```

peter-edge:
Ping

PR #109, opened 2018-06-19
Move internal/x/extract to internal/extract

peter-edge:


internal/extract/getter.go
```diff
@@ -0,0 +1,224 @@
...
+	if len(split) < 2 {
+		return nil, fmt.Errorf("no field for path %s", path)
+	}
+	message, err := g.GetMessage(fileDescriptorSets, strings.Join(split[0:len(split)-1], "."))
+	if err != nil {
+		return nil, err
+	}
+	var foundFieldDescriptorProto *descriptor.FieldDescriptorProto
+	for _, fieldDescriptorProto := range append(message.GetField(), message.GetExtension()...) {
+		if fieldDescriptorProto.GetName() == split[len(split)-1] {
```
AlexAPB: nit: maybe capture `split[len(split)-1]` as `wantName` or something for better readability

peter-edge: Sure why not :-)

PR #113, opened 2018-06-20
Add --update-file-options flag to format

peter-edge:
This adds a `--update-file-options` flag to `prototool format` and `prototool all` to update the values of `go_package, java_package` to the correct values based on the package per the style guide, and to delete `java_multiple_files` and `java_outer_classname`.


internal/cmd/flags.go
```diff
@@ -116,3 +117,7 @@ func (f *flags) bindProtocURL(flagSet *pflag.FlagSet) {
 func (f *flags) bindUncomment(flagSet *pflag.FlagSet) {
 	flagSet.BoolVar(&f.uncomment, "uncomment", false, "Uncomment the example config settings.")
 }
+
+func (f *flags) bindUpdateFileOptions(flagSet *pflag.FlagSet) {
+	flagSet.BoolVar(&f.updateFileOptions, "update-file-options", false, "Update the file options go_package, java_package to match the package per the guidelines of the style guide.")
```
AlexAPB: nit: replace comma with `and`

internal/format/first_pass_visitor.go
```diff
@@ -39,10 +40,14 @@ type firstPassVisitor struct {
 	Imports []*proto.Import
 
 	haveHitNonComment bool
+
+	updateFileOptions bool
```
AlexAPB: nit: Maybe `fixFileOptions` or `repairFileOptions`?

internal/cmd/flags.go
peter-edge: Done

internal/format/first_pass_visitor.go
peter-edge: I want this to match the flag

PR #116, opened 2018-06-20
Move internal/x/reflect to internal/reflect

peter-edge:


internal/reflect/handler.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func newHandler(options ...HandlerOption) *handler {
+	handler := &handler{
+		logger: zap.NewNop(),
+	}
+	for _, option := range options {
+		option(handler)
+	}
+	// TODO(pedge): composition
```
AlexAPB: can we drop this todo?

peter-edge: We can, but I still want to do it...so... :-)

AlexAPB: haha, then let's do it?!

peter-edge: I need to do this in like three places...perhaps this should be `// TODO: composition?` like as in a question lol

AlexAPB: lol sounds good, let's merge it

PR #118, opened 2018-06-21
Switch java_package, java_multiple_files, java_outer_classname to match Google Cloud API format

### LINKED ISSUE #117: Switch java_package, java_multiple_files, java_outer_classname to match Google Cloud API format
https://cloud.google.com/apis/design/file_structure

Specifically:

- `java_package` will be `com.PACKAGE` instead of `com.PACKAGE.pb`
- `java_multiple_files` will be set to true
- `java_outer_classname` will be `FILENAMEProto`

`go_package` will not change as there's no guidance on that, and we won't do anything with `csharp_namespace`, `objc_class_prefix`, or `php_namespace` as of yet, as we don't use those and we don't need to overcomplicate this for now (we might want to do `objc_class_prefix` in the future though).

Changes that need to be made:

- Update the linters to enforce this
- Update the create package to create Protobuf files with this
- Update the formatter to set these values after #113 is merged

peter-edge:
Fixes #117. See that issue for what amounts to the whole description of this PR.


internal/lint/check_file_options_equal.go
```diff
@@ -117,3 +125,9 @@ func packageBasename(pkg string) string {
 	split := strings.Split(pkg, ".")
 	return split[len(split)-1]
 }
+
+func fileBasenameUpperCamelCase(filename string) string {
```
AlexAPB: We have this same function in `lint` and `format`. We're also doing a bit of adhoc java/go package naming (ie `"foo" +"pb"), thoughts on creating/moving helper functions into the `strings` package?

peter-edge: I'm going to do a follow-up that cleans this up, is that OK?

AlexAPB: yup sounds good!

PR #119, opened 2018-06-21
 Switch --update-file-options to --no-update-file-options

peter-edge:
As opposed to opting in to `--update-file-options`, I think we might want it as the default, changing the flag to the negating `--no-update-file-options` for the current behavior. I only updated this in the `internal/cmd` package, doing `!flags.noUpdateFileOptions` when this is passed to the `internal/x/exec` package. If we don't get complaints and end up wanting this, we can go back and change everything from `updateFileOptions` to `noUpdateFileOptions` down the stack later, but I'd prefer not to have thrashing here, and also we might just want to keep it this way (it makes more sense logically in my head at the moment).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,13 +37,13 @@ golden: install
 		rm -f $${file}; \
 	done
 	for file in $(shell find internal/cmd/testdata/format -name '*.proto'); do \
-		prototool format $${file} > $${file}.golden || true; \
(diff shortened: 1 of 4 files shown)


peter-edge:
I'd prefer just to keep it as-is for now, we can change it later if we end up keeping this for 1.0.

PR #121, opened 2018-06-21
Move internal/x/grpc to internal/grpc

peter-edge:


internal/grpc/handler.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	fileDescriptorSet, err := desc.SortFileDescriptorSet(service.FileDescriptorSet, service.FileDescriptorProto)
+	if err != nil {
+		return nil, err
+	}
+	return grpcurl.DescriptorSourceFromFileDescriptorSet(fileDescriptorSet)
+}
+
+func getServiceForMethod(method string) (string, error) {
+	split := strings.Split(method, "/")
```
AlexAPB: Is `package.service/Method` a gRPC standard or just a `grpcurl` standard? If it's just `grpcurl` how would you feel if we separated these into two fields instead? `... package.service Method ...`

internal/grpc/invocation_event_handler.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/jhump/protoreflect/desc"
+	"go.uber.org/zap"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+var jsonpbMarshaler = &jsonpb.Marshaler{Indent: "  "}
+
+type invocationEventHandler struct {
```
AlexAPB: Can we add a static check with the `grpcurl.InvocationEventHandler` or a comment to indicate that we're implementing the below methods to match that interface.

internal/grpc/invocation_event_handler.go
```diff
@@ -0,0 +1,88 @@
...
+func (i *invocationEventHandler) OnReceiveResponse(message proto.Message) {
+	i.println(i.marshal(message))
+}
+
+func (i *invocationEventHandler) OnReceiveTrailers(s *status.Status, _ metadata.MD) {
+	if err := s.Err(); err != nil {
+		// TODO(pedge): not great for streaming
+		i.err = err
+		// printed by returning the error in handler
+		//i.println(err.Error())
```
AlexAPB: drop the comment

internal/grpc/invocation_event_handler.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (i *invocationEventHandler) OnReceiveHeaders(metadata.MD) {}
+
+func (i *invocationEventHandler) OnReceiveResponse(message proto.Message) {
+	i.println(i.marshal(message))
+}
+
+func (i *invocationEventHandler) OnReceiveTrailers(s *status.Status, _ metadata.MD) {
+	if err := s.Err(); err != nil {
+		// TODO(pedge): not great for streaming
```
AlexAPB: why not :)

internal/grpc/handler.go
peter-edge: It’s a gRPC standard, the request url is /package.service/method

internal/grpc/invocation_event_handler.go
peter-edge: Done

internal/grpc/invocation_event_handler.go
peter-edge: Done

internal/grpc/invocation_event_handler.go
peter-edge: Yea I guess it's fine

PR #123, opened 2018-06-22
Rename --no-update-file-options to --no-rewrite

peter-edge:
We might want to rewrite things other than file options in the future (for example, syntax and package).


lucasvo:
Is there any reason why you decided not to use a config option for this but a CLI flag? This to me sounds like it is something that should be configurable.

peter-edge:
Hmm that’s a good point. Need to think about that more, thanks for raising this @apeatsbond

PR #124, opened 2018-06-22
Centralize Protobuf package/filename string manipulation functionality

peter-edge:
This combines the string manipulation functionality to go from package and filename to `go_package`, `java_package`, and `java_outer_classname`.

### DIFF
--- a/internal/create/handler.go
+++ b/internal/create/handler.go
@@ -30,8 +30,8 @@ import (
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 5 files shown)


peter-edge:
I'd prefer to keep it in a separate package - I know it seems like overkill, maybe it is, but `strs` just contains project-independent string stuff, similar to https://golang.org/pkg/strings/, and this is for Protobuf string stuff.

PR #130, opened 2018-06-26
Add linter to verify no enums use the allow_alias option

### LINKED ISSUE #38: Handle allow_alias in enums
We probably want to outlaw enum aliases as part of linting by default, but this is a discussion item https://developers.google.com/protocol-buffers/docs/proto3#enum

peter-edge:
Fixes #38.

### DIFF
--- a/internal/lint/check_enums_no_allow_alias.go
+++ b/internal/lint/check_enums_no_allow_alias.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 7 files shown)


internal/lint/check_enums_no_allow_alias.go
```diff
@@ -0,0 +1,66 @@
...
+		child.Accept(v)
+	}
+}
+
+func (v enumsNoAllowAliasVisitor) VisitOption(element *proto.Option) {
+	// file option
+	if element.Parent == nil {
+		return
+	}
+	switch element.Parent.(type) {
```
AlexAPB: nit: maybe just an `if-ok` instead of a switch since there's only ever one case, I don't really care either way

internal/lint/check_enums_no_allow_alias.go
```diff
@@ -0,0 +1,66 @@
...
+	case (*proto.Enum):
+		if element.Name == "allow_alias" {
```
AlexAPB: shouldn't we check the value? Some people may have `allow_alias = false;` which I'll admit is redundant, but still valid.

internal/lint/check_enums_no_allow_alias.go
peter-edge: Eh for this type statement I like this a bit more if that's OK, it's a little cleaner potentially.

internal/lint/check_enums_no_allow_alias.go
peter-edge: I don't want people to specify `allow_alias = false` either, I'm linting that the value isn't set at all, I did that with `java_multiple_files` and `java_outer_classname` before we changed everything around too.

PR #135, opened 2018-07-05
Add dry-run flag to replace protoc-commands

amckinney:
This replaces the `protoc-commands` command with a `--dry-run` flag in the relevant commands: compile, format, gen, and lint. Using a `dry-run` flag is featured in a variety of other CLI tools, such as git, and it removes the need for an additional command.

Previously, we only supported `protoc-commands` for `compile` and `gen`. It was simple enough to introduce these to `format` and `lint`. I'm not opposed to removing this functionality, as it might not be as necessary for `format` and `lint`, but thought it was worth including here.

### DIFF
--- a/internal/exec/runner.go
+++ b/internal/exec/runner.go
@@ -183,23 +183,23 @@ func (r *runner) Files(args []string) error {
 	return nil
 }
 
-func (r *runner) Compile(args []string) error {
+func (r *runner) Compile(args []string, dryRun bool) error {
 	meta, err := r.getMeta(args)
 	if err != nil {
 		return err
 	}
 	r.printAffectedFiles(meta)
-	_, err = r.compile(false, false, meta)
+	_, err = r.compile(false, false, dryRun, meta)
 	return err
 }
 
-func (r *runner) Gen(args []string) error {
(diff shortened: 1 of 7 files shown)


internal/exec/runner.go
```diff
@@ -326,13 +324,13 @@ func (r *runner) ProtocCommands(args []string, genCommands bool) error {
...
 
-func (r *runner) Lint(args []string) error {
+func (r *runner) Lint(args []string, dryRun bool) error {
 	meta, err := r.getMeta(args)
 	if err != nil {
 		return err
 	}
 	r.printAffectedFiles(meta)
-	if _, err := r.compile(false, false, meta); err != nil {
+	if _, err := r.compile(false, false, dryRun, meta); err != nil {
```
peter-edge: `dry-run` doesn't really apply for `lint` and `format` - they're just going to print out the commands from `compile`, it's not printing out the dry run of what it's linting, which will be confusing (it will just print the `protoc` commands from `compile`, but then does more stuff). I would remove for `lint` and `format`.

amckinney: Yea, sounds reasonable. I'll update!

PR #137, opened 2018-07-06
Collape list commands into lint flags

amckinney:
Similar to https://github.com/uber/prototool/pull/135, this collapses the `lint-*` commands into CLI flags for `prototool lint`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - The protoc-commands command is now accessible via the `dry-run`
   flag for compile, format, gen, and lint.
+- The list-* commands are now accessible via lint command flags:
+  list-all, list-configured, list-groups, and list-group.
 
 ## [0.4.0] - 2018-06-22
 ### Added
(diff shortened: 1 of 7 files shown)


amckinney:
Seeing as all of these list commands concern themselves with how `prototool lint` will behave, namely what linters will be used and/or are available to be configured, I think it makes more sense to include them as flags. With this, users can clearly see what flags are unique to the lint command. 

For example,

```
$ prototool lint -h
Lint proto files and compile with protoc to check for failures.

Usage:
  prototool lint dirOrProtoFiles... [flags]

Flags:
      --dir-mode            Run as if the directory the file was given, but only print the errors from the file. Useful for integration with editors.
  -h, --help                help for lint
      --list-all            List all of the available linters.
      --list-configured     List all of the configured linters.
      --list-group string   List all of the linters configured in the given group.
      --list-groups         List all of the available linter groups.
```

To me, this is much easier to work with and eliminates the need to jump between multiple top-level commands when such commands only pertain to the behavior of one other specifically (i.e. the relationship between `prototool list-all-lint-groups` and `prototool lint`). My assumed thought process from a user would be: "I want to lint my proto files. What options do I have access to? This is probably retrievable by accessing that command's particular help output. Let's try `prototool lint -h`"

What are your thoughts? cc @peter-edge @AlexAPB

AlexAPB:
I see both sides of this:
- minimizing the surface area for 1.0
- not overloading commands

I do think that it's verbose to have four commands that all more or less do the same thing. A middle ground could be having a new `list-linters`(or whatever) command that consolidates these. However, I think I am comfortable with @amckinney suggested changes as the help menu helps clarify the use-cases.

Let's discuss this further offline.

peter-edge:
Making these into flags does not minimize the surface area for v1.0 - flags still have to be supported, it's the same thing. I guess what I'm missing is the difference between sub-commands and flags in terms of verbosity - to me, introducing flags that do not have an effect on the main action is more confusing.

peter-edge:
Superseded by #143.

Issue #131, opened 2018-06-26
Add tool to check for breaking changes


peter-edge:
https://github.com/nilslice/protolock just came out recently, and this is a nice tool, but it would be nice if similar functionality was built into Prototool. Additionally, Protolock generates a `proto.lock` file that stores information on the Protobuf files in a repository, which I'd like to get rid of. This is a new format and inherently can have bugs, but more to the point is duplicated information - the `proto.lock` file is generated via the checked-in Protobuf files, which in turn contain all the information you need. If you were to generate a `FileDescriptorSet` for the current Protobuf files, and a `FileDescriptorSet` for the old Protobuf files, and then compare the two, you could get the same thing. A `proto.lock` file could have just been the result of `--descriptor_set_out` as a further example, and then used https://github.com/jhump/protoreflect, but again instead of storing another file, we can just generate this on the fly. I'd also like errors to be printed in the Prototool standard `filename:line:column:message` format.

Example of what I'd imagine:

```
prototool check-compatibility --from dev # dev is the name of the base branch
a/b/c.proto:3:4:Cannot remove field "foo".
a/b/d.proto:1:1:Cannot change syntax.
...
```

AlexAPB:
Related: #219

peter-edge:
This is done in https://github.com/uber/prototool/pull/279. You can use this on dev:

```
prototool break check --git-tag TAG # check against this tag
prototool break check --git-branch BRANCH # check against this branch
prototool break check path/to/protodir --git tag TAG # check against this tag for subdir path/to/protodir
prototool break check # check against the default branch
prototool break check path/to/protodir # check against the default branch for subdir path/to/protodir
```

More documentation to follow. Packages of the form `foo.bar.vMAJORbetaBETA` where MAJOR >=1 and BETA >= 1 are ignored unless the `--include-beta` flag is added.

peter-edge:
See https://github.com/uber/prototool/pull/382 for additional workflows

Issue #93, opened 2018-05-25
Multiple prototool.yaml files in the same project


amckinney:
We should consider the `prototool.yaml` hierarchy and determine whether or not we should stick to the current tree/override behavior. A single `prototool.yaml` at the project's root might be sufficient for now. This would also simplify some of the implementation details found in the `settings` package. 

If there isn't an obvious need for the hierarchy, it might be worth deferring until after a 1.0 release.

AlexAPB:
I'm in favour of one `prototool.yaml` for one project and slimming down `internal/settings`. If there is a demand for this, we can reintroduce post `1.0`. Let's not make this a `1.0` blocker, however.

peter-edge:
I think you'll want this too, but note moving to this now will lock you in later. The big thing is everywhere you see `protoSets ...*file.ProtoSet` and `fileDescriptorSets []*FileDescriptorSet` - as the code is now, it supports multiple sets of Protocol Buffer files (corresponding to multiple `prototool.yaml` files), and deleting that logic will make it complicated to add it back in the future (it was complicated to move to in the first place).

If you want to lock it in, that's fine too, we probably only will want a single `prototool.yaml` file, but just an FYI.

Note this issue is more or less a duplicate of https://github.com/uber/prototool/issues/10.

peter-edge:
Closing this for now as we went with one `prototool.yaml` file, we can re-visit in the future if we need a different setup.

Issue #10, opened 2018-04-10
Config file discovery semantics


peter-edge:
Right now, config files are named `prototool.yaml` and searched for starting at the Protobuf file path, and going up a directory until hitting root. All paths in the config files can be relative or absolute, but if relative, the actual path used will start at the config file directory. If using directory mode, this means multiple `prototool.yaml` files corresponding to multiple found directories with Protobuf files may be used. After a config file is found, Prototool walks starting at that directory searching for Protobuf files, or if no config file is found, Prototool searches starting at the current directory.

A few things:

- Is `prototool.yaml` the right name?
- Should we terminate earlier? Right now for example, if you don't have a `prototool.yaml` file in your repo, but do in your home directory, `prototool` will go all the way down to your home directory, and then walk everything in your home directory, which can be annoying.
- Should we require a `prototool.yaml` file to use `prototool`? See the above. It's good practice to always have a `prototool.yaml` file to at least lock down the version of `protoc` you use, but this can be annoying for OSS users who are just starting out.
- **The biggest thing:** should we not allow multiple `prototool.yaml` files to be discovered? If multiple `prototool.yaml` files are discovered, this means that multiple calls to `protoc` must be done as they denote different builds, and more than one `FileDescriptorSet` is created. This gets bad when you do e.g. gRPC calls as `foo.bar.BazService` could be in more than one `FileDescriptorSet`, so you basically have to stop when you find the first one. All packages in Prototool have to handle multiple `FileDescriptorSets`, which leads to some other bad outcomes. Most people shouldn't have multiple `prototool.yaml` files, but it could be useful for compiling vendored `prototool.yaml` files.

amckinney:
I think that requiring a `prototool.yaml` file is fine / necessary. It determines the root for how a directory's underlying proto files are treated (e.g. how they're linted / generated), and it clearly expresses a project's protoc plugins, linters, etc.

I see where you're coming from with respect to on-boarding, but adding a configuration file to begin using a new tool seems to be a reasonable requirement. And now that we are only allowing one `prototool.yaml`, this is even easier to justify. 

The naming seems fine to me as well. The tool is called `prototool` - we may as well name its configuration file the same. An alternative would be to house an arbitrarily-named file in a `.prototool/` directory, but that doesn't have much benefit.

What are your thoughts? I'm fine with resolving this issue for now if others agree. cc @peter-edge @AlexAPB

peter-edge:
[...]

PR #139, opened 2018-07-09
Allow only one prototool.yaml file

peter-edge:
Per #10 and #93, this change does the bare minimum to enforce only one `prototool.yaml` file per command. There is further refactoring that can be done after this, but this propagates the changes through the codebase while retaining the multiple `prototool.yaml` file support in the implementation for now. Putting this up for discussion, it also helps simplify the logic for #131.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,35 +88,7 @@ The command `prototool init` will generate a config file in the current director
 
 When specifying a directory or set of files for Prototool to operate on, Prototool will search for config files for each directory starting at the given path, and going up a directory until hitting root. If no config file is found, Prototool will use default values and operate as if there was a config file in the current directory, including the current directory with `-I` to `protoc`.
 
-While almost all projects should not have multiple `prototool.yaml` files (and this [may be enforced before v1.0](https://github.com/uber/prototool/issues/10)), as of now, multiple `prototool.yaml` files corresponding to multiple found directories with Protobuf files may be used. For example, if you have the following layout:
-
-```
-.
-├── a
-│   ├── d
-│   │   ├── file.proto
-│   │   ├── file2.proto
-│   │   ├── file3.proto
-│   │   └── prototool.yaml
-│   ├── e
-│   │   └── file.proto
-│   ├── f
-│   │   └── file.proto
-│   └── file.proto
-├── b
-│   ├── file.proto
-│   ├── g
-│   │   └── h
-│   │       └── file.proto
-│   └── prototool.yaml
-├── c
-│   ├── file.proto
-│   └── i
-│       └── file.proto
-└── prototool.yaml
-```
-
-Everything under `a/d` will use `a/d/prototool.yaml`, everything under `b`, `b/g/h` will use `b/prototool.yaml`, and everything under `a`, `a/e`, `a/f`, `c`, `c/i` will use `prototool.yaml`. See [internal/file/testdata](internal/file/testdata) for the most current example.
+If multiple `prototool.yaml` files are found that match the input directory or files, an error will be returned. We have an ongoing discussion about whether to allow multiple `prototool.yaml` files, see [this issue](https://github.com/uber/prototool/issues/10) for more details.
 
 ## File Discovery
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -88,35 +88,7 @@ The command `prototool init` will generate a config file in the current director
...
-│   └── prototool.yaml
-├── c
-│   ├── file.proto
-│   └── i
-│       └── file.proto
-└── prototool.yaml
-```
-
-Everything under `a/d` will use `a/d/prototool.yaml`, everything under `b`, `b/g/h` will use `b/prototool.yaml`, and everything under `a`, `a/e`, `a/f`, `c`, `c/i` will use `prototool.yaml`. See [internal/file/testdata](internal/file/testdata) for the most current example.
+If multiple `prototool.yaml` files are found that match the input directory or files, an error will be returned. We have an ongoing discussion about whether to allow multiple `prototool.yaml` files, see [this issue](https://github.com/uber/prototool/issues/10) for more details.
```
amckinney: What does it mean for `prototool.yaml` files to match the input directory or files? 

Seeing as you cannot have more than one `prototool.yaml` file in a given directory, could we simply just use the `prototool.yaml` file that corresponds to where the command was run, and ignore all others? For example, in your removed tree example, running `prototool <cmd>` in `b/g` should just use the `prototool.yaml` found there (and not the one included in the root). Similarly, running the same command at the root would use the root `prototool.yaml` and ignore all others.

Let me know if this isn't what you meant - just trying to determine when an error would occur and whether or not it's necessary.

internal/protoc/compiler.go
```diff
@@ -168,31 +164,27 @@ func (c *compiler) Compile(protoSets ...*file.ProtoSet) (*CompileResult, error)
...
-			return nil, err
-		}
-		for _, cmdMeta := range cmdMetas {
-			cmdMetaStrings = append(cmdMetaStrings, cmdMeta.String())
-		}
-		cleanCmdMetas(cmdMetas)
+func (c *compiler) ProtocCommands(protoSet *file.ProtoSet) ([]string, error) {
+	// we end up calling the logic that creates temporary files for file descriptor sets
+	// anyways, so we need to clean them up with cleanCmdMetas
+	// this logic could be simplified to have a "dry run" option, but ProtocCommands
```
amckinney: nit: Now that this uses the `dry-run` flag, could we remove this comment? My fault for not handling in https://github.com/uber/prototool/pull/135

peter-edge: Can't remove this yet, needs some refactoring, this still creates temporary files.

internal/file/proto_set_provider.go
```diff
@@ -52,7 +52,45 @@ func newProtoSetProvider(options ...ProtoSetProviderOption) *protoSetProvider {
...
+	case 0:
+		return nil, fmt.Errorf("no proto files found for dirPath %q", dirPath)
+	case 1:
+		return protoSets[0], nil
+	default:
+		configDirPaths := make([]string, 0, len(protoSets))
+		for _, protoSet := range protoSets {
+			configDirPaths = append(configDirPaths, protoSet.Config.DirPath)
+		}
+		return nil, fmt.Errorf("multiple configuration files found in directories %v for proto files found for dirPath %q but expected exactly one configuration file", configDirPaths, dirPath)
```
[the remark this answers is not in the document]
peter-edge: Done

internal/file/proto_set_provider.go
```diff
@@ -52,7 +52,45 @@ func newProtoSetProvider(options ...ProtoSetProviderOption) *protoSetProvider {
...
+	case 0:
+		return nil, fmt.Errorf("no proto files found for filePaths %v", filePaths)
+	case 1:
+		return protoSets[0], nil
+	default:
+		configDirPaths := make([]string, 0, len(protoSets))
+		for _, protoSet := range protoSets {
+			configDirPaths = append(configDirPaths, protoSet.Config.DirPath)
+		}
+		return nil, fmt.Errorf("multiple configuration files found in directories %v for proto files found for filePaths %v but expected exactly one configuration file", configDirPaths, filePaths)
```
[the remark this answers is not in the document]
peter-edge: Done

Issue #140, opened 2018-07-09
Functions binary-to-json and json-to-binary should take in a flag for data or stdin


peter-edge:
Similar to #97.

amckinney:
These commands have been unregistered - closing for now.

PR #141, opened 2018-07-09
Move address, method, and input data to flags for gRPC

### LINKED ISSUE #97: Behavior is confusing if you do not specify an argument for data for grpc
Example: `cat input.json | prototool grpc example 0.0.0.0:8080 foo.ExcitedService/Exclamation`

The trailing `-` is missing to denote stdin reading, but this is technically a legal invocation where it assumes you mean the current directory, with address `example`, method name `0.0.0.0:8080`, and data `foo.ExcitedService/Exclamation`. We need to fix this in some manner, perhaps flags for the input args.

peter-edge:
Fixes #97.

#140 will be a follow-up after this.

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -65,7 +65,7 @@ type Runner interface {
 	BinaryToJSON(args []string) error
 	JSONToBinary(args []string) error
 	All(args []string, disableFormat, disableLint, rewrite bool) error
-	GRPC(args, headers []string, callTimeout, connectTimeout, keepaliveTime string) error
+	GRPC(args []string, address string, method string, data string, stdin bool, headers []string, callTimeout, connectTimeout, keepaliveTime string) error
 }
 
 // RunnerOption is an option for a new Runner.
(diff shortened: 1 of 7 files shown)


internal/exec/exec.go
```diff
@@ -65,7 +65,7 @@ type Runner interface {
 	BinaryToJSON(args []string) error
 	JSONToBinary(args []string) error
 	All(args []string, disableFormat, disableLint, rewrite bool) error
-	GRPC(args, headers []string, callTimeout, connectTimeout, keepaliveTime string) error
+	GRPC(args []string, address string, method string, data string, stdin bool, headers []string, callTimeout, connectTimeout, keepaliveTime string) error
```
amckinney: nit: Can we keep the arguments grouped by type?
```
GRPC(args, headers []string, address, method, data, callTimeout, connectTimeout, keepaliveTime string, stdin bool) error
```

peter-edge: Done

PR #143, opened 2018-07-10
Delete commands, flags, options in preparation for v1.0

peter-edge:
Per our discussion, this:

- Deletes the commands `binary-to-json, clean, descriptor-proto, download, field-descriptor-proto, json-to-binary, list-all-linters, list-all-lint-groups, list-linters, list-lint-group, service-descriptor-proto`.
- Adds the functionality for the commands `list-all-linters, list-linters` to flags under `lint`.
- Deletes the global flags `--cache-path, --print-fields`.
- Moves the flag `--dry-run` to just be scoped to the `compile, gen` commands (I missed this in #135).
- Undocuments the option `lint.group` in the `prototool.yaml` configuration. This is still supported under the hood as there's a key test that uses the `all` group to test against all linters in the codebase, but it's not documented and therefore we make no claims to support it.
- Does some cleanup in the README.md.

We may want to reconsider keeping the `download` and maybe `clean` commands, as I think some use this to seed their CI process, and it's really useful to just go to the cache directory easily.

The old help page for `prototool`:

```
$ prototool -h
Usage:
  prototool [command]

Available Commands:
  all                      Compile, then format and overwrite, then re-compile and generate, then lint, stopping if any step fails.
  binary-to-json           Convert the data from json to binary for the message path and data.
  clean                    Delete the cache.
  compile                  Compile with protoc to check for failures.
  create                   Create the given Protobuf files according to a template that passes default prototool lint.
  descriptor-proto         Get the descriptor proto for the message path.
  download                 Download the protobuf artifacts to a cache.
  field-descriptor-proto   Get the field descriptor proto for the field path.
  files                    Print all files that match the input arguments.
  format                   Format a proto file and compile with protoc to check for failures.
  gen                      Generate with protoc.
  grpc                     Call a gRPC endpoint. Be sure to set required flags address, method, and either data or stdin.
  help                     Help about any command
  init                     Generate an initial config file in the current or given directory.
  json-to-binary           Convert the data from json to binary for the message path and data.
  lint                     Lint proto files and compile with protoc to check for failures.
  list-all-lint-groups     List all the available lint groups.
  list-all-linters         List all available linters.
  list-lint-group          List the linters in the given lint group.
  list-linters             List the configurerd linters.
  service-descriptor-proto Get the service descriptor proto for the service path.
  version                  Print the version.

Flags:
      --cache-path string     The path to use for the cache, otherwise uses the default behavior.
      --debug                 Run in debug mode, which will print out debug logging.
      --dry-run               Print the protoc commands that would have been run without actually running them.
      --harbormaster          Print failures in JSON compatible with the Harbormaster API.
  -h, --help                  help for prototool
      --print-fields string   The colon-separated fields to print out on error. (default "filename:line:column:message")
      --protoc-url string     The url to use to download the protoc zip file, otherwise uses GitHub Releases. Setting this option will ignore the config protoc_version setting.

Use "prototool [command] --help" for more information about a command.
```

The new help page for `prototool`:

```
$ prototool -h
Usage:
  prototool [command]

Available Commands:
  all         Compile, then format and overwrite, then re-compile and generate, then lint, stopping if any step fails.
  compile     Compile with protoc to check for failures.
  create      Create the given Protobuf files according to a template that passes default prototool lint.
  files       Print all files that match the input arguments.
  format      Format a proto file and compile with protoc to check for failures.
  gen         Generate with protoc.
  grpc        Call a gRPC endpoint. Be sure to set required flags address, method, and either data or stdin.
  help        Help about any command
  init        Generate an initial config file in the current or given directory.
  lint        Lint proto files and compile with protoc to check for failures.
  version     Print the version.

Flags:
      --debug               Run in debug mode, which will print out debug logging.
      --harbormaster        Print failures in JSON compatible with the Harbormaster API.
  -h, --help                help for prototool
      --protoc-url string   The url to use to download the protoc zip file, otherwise uses GitHub Releases. Setting this option will ignore the config protoc_version setting.

Use "prototool [command] --help" for more information about a command.
```

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -50,22 +50,26 @@ var genManTime = time.Date(2018, time.January, 1, 0, 0, 0, 0, time.UTC)
 
 // Do runs the command logic.
 func Do(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int {
-	return runRootCommand(args, stdin, stdout, stderr, (*cobra.Command).Execute)
+	return do(false, args, stdin, stdout, stderr)
+}
+
+func do(develMode bool, args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int {
+	return runRootCommand(develMode, args, stdin, stdout, stderr, (*cobra.Command).Execute)
 }
 
 // GenBashCompletion generates a bash completion file to the writer.
 func GenBashCompletion(stdin io.Reader, stdout io.Writer, stderr io.Writer) int {
-	return runRootCommandOutput([]string{}, stdin, stdout, stderr, (*cobra.Command).GenBashCompletion)
+	return runRootCommandOutput(false, []string{}, stdin, stdout, stderr, (*cobra.Command).GenBashCompletion)
 }
 
 // GenZshCompletion generates a zsh completion file to the writer.
 func GenZshCompletion(stdin io.Reader, stdout io.Writer, stderr io.Writer) int {
-	return runRootCommandOutput([]string{}, stdin, stdout, stderr, (*cobra.Command).GenZshCompletion)
+	return runRootCommandOutput(false, []string{}, stdin, stdout, stderr, (*cobra.Command).GenZshCompletion)
 }
 
 // GenManpages generates the manpages to the given directory.
 func GenManpages(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int {
-	return runRootCommand(args, stdin, stdout, stderr, func(cmd *cobra.Command) error {
+	return runRootCommand(false, args, stdin, stdout, stderr, func(cmd *cobra.Command) error {
 		if len(args) != 1 {
 			return fmt.Errorf("usage: %s dirPath", os.Args[0])
 		}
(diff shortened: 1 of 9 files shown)


internal/cmd/cmd.go
```diff
@@ -78,21 +82,23 @@ func GenManpages(args []string, stdin io.Reader, stdout io.Writer, stderr io.Wri
 	})
 }
 
-func runRootCommandOutput(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer, f func(*cobra.Command, io.Writer) error) int {
-	return runRootCommand(args, stdin, stdout, stderr, func(cmd *cobra.Command) error { return f(cmd, stdout) })
+// develMode turns on sub-commands and potentially flags that we do not expose during the build of the prototool binary
```
amckinney: nit: missing period

internal/cmd/cmd.go
```diff
@@ -78,21 +82,23 @@ func GenManpages(args []string, stdin io.Reader, stdout io.Writer, stderr io.Wri
...
+func runRootCommandOutput(develMode bool, args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer, f func(*cobra.Command, io.Writer) error) int {
+	return runRootCommand(develMode, args, stdin, stdout, stderr, func(cmd *cobra.Command) error { return f(cmd, stdout) })
 }
 
-func runRootCommand(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer, f func(*cobra.Command) error) (exitCode int) {
+// develMode turns on sub-commands and potentially flags that we do not expose during the build of the prototool binary
```
amckinney: nit: missing period.

internal/exec/exec.go
```diff
@@ -56,9 +56,7 @@ type Runner interface {
 	DescriptorProto(args []string) error
 	FieldDescriptorProto(args []string) error
 	ServiceDescriptorProto(args []string) error
-	Lint(args []string) error
-	ListLinters() error
-	ListAllLinters() error
+	Lint(args []string, listAllLinters bool, listLinters bool) error
```
amckinney: nit: `Lint(args []string, listAllLinters, listLinters bool) error`

internal/cmd/cmd.go
peter-edge: I don't put periods at the end of short comments, this is all over the codebase lol, this is just instructive

internal/cmd/cmd.go
peter-edge: Same

internal/exec/exec.go
peter-edge: This actually is not the style of Prototool - I think it leads to serious errors that Golang does this, and I always specify fully. I did this in the GRPC function, but if it's coming up multiple times we should talk about it, and if you feel strongly, this should be consistent codebase-wide (where right now the only time the shorthand is used in the entire codebase is on GRPC).

peter-edge:
@amckinney if you deregister the harbormaster flag, this will not work with our internal setup at all, which probably defeats the whole purpose.

internal/cmd/cmd.go
amckinney: Ok.

internal/exec/exec.go
amckinney: Sure, works for me.

Issue #145, opened 2018-07-10
Move cmd/* to internal


peter-edge:
I want to make it explicit that you should not be `go get`'ing this since we don't have `vendor` checked in. Thoughts @AlexAPB @amckinney?

PR #147, opened 2018-07-10
Move cmd/prototool to internal/cmd/prototool

peter-edge:
See #145.


README.md
```diff
@@ -50,6 +50,11 @@ curl -sSL https://github.com/uber/prototool/releases/download/v0.4.0/prototool-$
   chmod +x /usr/local/bin/prototool
 ```
 
+Prototool is purposefully all put under internal in an attempt to emphasize that you should not install `prototool` with `go get`.
+Although you can technically `go get` from `internal`, we do not check in the vendor directory, so `go get` will not respect the
+versions in `glide.yaml`. We have specific version requirements, so not using these can and probably will result in errors building
```
amckinney: nit: versions in `glide.yaml`. We have specific version requirements, so not using these can, and probably will, result in errors **when** building ...

peter-edge: Done

PR #148, opened 2018-07-10
Update RELEASE.md

bufdev:
This has been on my TODO list for a while. This:

- Adds a note that you need to update the version in the Installation section of the README.md and updates the `git` commands appopriately.
- Removes the `sed` command to change the version. You should do this manually in the README.md, and the `sed` command always made me a little queasy.
- Adds a missing period.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -44,17 +44,18 @@ This document outlines how to create a release of prototool.
     ```
 
 
-5.  Update the version number in internal/vars/vars.go and verify that it
-    matches what is in the changelog.
+5.  Update the version number in internal/vars/vars.go and the Installation
+    section of the README.md and verify that it matches what is in the changelog.
 
-    ```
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -44,17 +44,18 @@ This document outlines how to create a release of prototool.
     ```
 
 
-5.  Update the version number in internal/vars/vars.go and verify that it
-    matches what is in the changelog.
+5.  Update the version number in internal/vars/vars.go and the Installation
```
amckinney: nit: Looks like you're missing a word here. Consider (or something similar):
```
Update the version number in internal.vars/vars.go, and verify that the changelog matches the version found in the Installation section of the README.md.
```

peter-edge: What am i missing? This reads fine to me. This is also the RELEASE.md doc, we don't need prose :-)

Issue #146, opened 2018-07-10
Flesh out help pages for commands


peter-edge:
We can probably copy/paste a lot from the README.md.

PR #149, opened 2018-07-11
Add cmdTemplate to build cobra.Commands

peter-edge:
I started out wanting to do #146, and wanted to move the documentation strings to somewhere static. I then realized it would be nice if we refactored `internal/cmd/cmd.go` to have the construction of `*cobra.Commands` use common functionality. I came up with `cmdTemplate`, which builds `*cobra.Commands` and moves the static parts of the `*cobra.Command` construction outside of `internal/cmd/cmd.go`. `cmdTemplates` allow us to:

- Take care of the copy/paste logic of `checkCmd` and just specify the `Run` portion of a `*cobra.Command` in terms of `func(exec.Runner, []string, *flags) error`.
- Prepend the `Short` string to the `Long` string when I add the `Long` strings for #146.
- Generally clean up the `internal/cmd` package so it's more accessible.

I also made the error message from `checkOS()` less snarky, I added this message during the prototyping phase.

The logic here is largely copy/paste, there's more cleanup to do I'm sure but this is a good first step.

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -38,8 +38,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
 	"github.com/uber/prototool/internal/exec"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 )
 
 // when generating man pages, the current date is used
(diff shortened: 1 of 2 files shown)


peter-edge:
Ping @amckinney @AlexAPB

internal/cmd/templates.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	command.Args = c.Args
+	command.Run = func(_ *cobra.Command, args []string) {
+		checkCmd(exitCodeAddr, stdin, stdout, stderr, args, flags, c.Run)
+	}
+	if c.BindFlags != nil {
+		c.BindFlags(command.PersistentFlags(), flags)
+	}
+	return command
+}
```
amckinney: Nice! If you like this more, you could map the direct field translations at the time of the `*cobra.Command` construction:
```
command := &cobra.Command{
  Args:   c.Args,
  Short:  c.Short,
  Use:    c.Use,
  Run:    func(_ *cobra.Command, args []string) { ... },
}
...
```

internal/cmd/templates.go
```diff
@@ -0,0 +1,381 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/uber/prototool/internal/exec"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+var (
+	allCmdTemplate = &cmdTemplate{
```
amckinney: Seeing as we only ever write out the `<cmd>Template` and call `.Build`, we could leave all of these named `allCmd`, `binaryToJSONCmd`, etc. and drop the `Template` suffix.

I'll leave it up to you.

internal/cmd/cmd.go
```diff
@@ -349,80 +144,8 @@ func checkOS() error {
...
-		return nil, err
-	}
-	return exec.NewRunner(workDirPath, stdin, stdout, runnerOptions...), nil
-}
-
-func getLogger(stderr io.Writer, debug bool) (*zap.Logger, error) {
-	level := zapcore.InfoLevel
-	if debug {
-		level = zapcore.DebugLevel
+		return fmt.Errorf("%s is not a supported operating system", runtime.GOOS)
```
amckinney: Nice 👍

internal/cmd/templates.go
peter-edge: I'd prefer these be explicitly named - they aren't `Commands`, they're `cmdTemplates`.

internal/cmd/templates.go
peter-edge: I liked running down the fields in order if that's OK?

amckinney: Yup, fine with me!

internal/cmd/templates.go
amckinney: 👍

Issue #16, opened 2018-04-10
Command-line file/directory specification


peter-edge:
Right now, most commands take something along the lines of `dirOrProtoFiles...` as their first arguments. These commands also have a flag `--dir-mode`. The possibilities:

- You can specify multiple files. If this is done, these files will be explicitly used for `protoc` calls etc.
- You can specify exactly one directory. If this is done, Prototool goes up until it finds a `prototool.yaml` file (or uses the current directory if none is found), and then walks starting at this location for all `.proto` files, and these are used, except for files in the `excludes` lists in `prototool.yaml` files.
- You can specify exactly one file, along with `--dir-mode`. This has the effect as if you specified the directory of this file (using the logic above), but errors are only printed for that file. This is useful for e.g. Vim integration.

The idea with "directory builds" is that you often need more than just one file to do a `protoc` call, for example if you have types in other files in the same package that are not referenced by their fully-qualified name, and/or if you need to know what directories to specify with `-I` to `protoc` (by default, the directory of the `prototool.yaml` file is used).

In general practice, directory builds are what you always want to do. File builds were just added for convenience.

There's a proposal to do the following instead:

- Only one argument is specified. This has the side effect of making the number of arguments to many commands consistent.
- If a directory is speficied, do directory mode, as usual.
- If a file is specified, do the equivalent of `--dir-mode` above.

There are some side effects, for example you need to special-case the `format` command probably, but we probably want to do this. It's not as trivial as I first realized, and wanted to open this up for discussion before I do it.

amckinney:
I agree with the proposal here - effectively remove the `dir-mode` flag and update the behavior so that directories are always processed as directories and single files will honor the "Only use this file, but from the context of the directory that contains it." It will reduce the cognitive load of what `dir-mode` means in the different contexts.

Let's look into this and make sure it, or something similar, happens before 1.0.

peter-edge:
Note when this change is made, you'll have to make changes to the Vim plugin.

PR #152, opened 2018-07-16
Start adding long help documentation strings

### LINKED ISSUE #146: Flesh out help pages for commands

peter-edge:
This starts the work to add long documentation strings to commands. This does a few things:

- Cleans up the combination of the `Short` and `Long` parameters by adding an extra newline.
- Uses https://github.com/mitchellh/go-wordwrap to wrap the documentation to 80 characters. This repository is MIT-licensed, so it's acceptable to use, and this makes the documentation look much cleaner (I was slightly surprised to find out that Cobra doesn't have an option for this). Some lines have to be manually wrapped to work, namely things in lists (so the next line is indented), and commands (where I `\-` them out to separate lines when displayed). All the long documentation strings look proper when I try them out.
- Copies over some of the documentation from README.md.

After this start, we can fill out the long documentation strings for other commands at will.

I'd recommend checking out this branch, doing `make install`, and then doing `prototool help grpc, lint,compile,...` to see how this looks.

@amckinney I know you have specific preferences for documentation, and it almost always just adds improvement, but my goal is to get this in as a start, and then iterate on the documentation. We can do one of a few things:

1. Just iterate on it on this PR like normal (which is fine if that's what you want to do! No issues!).
2. Land this, and have you do a new PR with any documentation changes you want.
3. You do a chained PR with any documentation changes.

Or...

4. You just push the changes you want directly to this branch/PR. This is probably the most efficient - instead of commenting the changes on the PR, just make the changes (which I'll almost for sure be fine with) and then we can land it.

(4) would probably be my preferred mechanism, but anything is fine, what do you think?

Fixes #146.

### DIFF
--- a/internal/cmd/templates.go
+++ b/internal/cmd/templates.go
@@ -221,6 +221,86 @@ If Vim integration is set up, files will be generated when you open a new Protob
 	grpcCmdTemplate = &cmdTemplate{
 		Use:   "grpc dirOrProtoFiles...",
 		Short: "Call a gRPC endpoint. Be sure to set required flags address, method, and either data or stdin.",
+		Long: `What this does behind the scenes:
+
+- Compiles your Protobuf files with "protoc", generating a "FileDescriptorSet".
+- Uses the "FileDescriptorSet" to figure out the request and response type for
+  the endpoint, and to convert the JSON input to binary.
+- Calls the gRPC endpoint.
+- Uses the "FileDescriptorSet" to convert the resulting binary back to JSON,
+  and prints it out for you.
+
+All these steps take on the order of milliseconds, for example the overhead for a file with four dependencies is about 30ms, so there is little overhead for CLI calls to gRPC.
+
+There is a full example for gRPC in the example directory of Prototool. Run "make init example" to make sure everything is installed and generated.
+
+Start the example server in a separate terminal by doing "go run example/cmd/excited/main.go".
+
+prototool grpc dirOrProtoFiles... \
+  --address serverAddress \
+  --method package.service/Method \
+  --data 'requestData'
+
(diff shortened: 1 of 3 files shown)


amckinney:
Nice! Yea, let's take option (4) and add commits to this PR. Once we have all of the commands documented, we can land them all together.

peter-edge:
Can you make this the next thing you work on regarding Prototool if possible? It'd be cool to close this out :-)

internal/cmd/templates.go
```diff
@@ -161,6 +221,86 @@ var (
 	grpcCmdTemplate = &cmdTemplate{
 		Use:   "grpc dirOrProtoFiles...",
 		Short: "Call a gRPC endpoint. Be sure to set required flags address, method, and either data or stdin.",
+		Long: `What this does behind the scenes:
+
+- Compiles your Protobuf files with "protoc", generating a "FileDescriptorSet".
```
AlexAPB: I think this part may be unnecessary, describing a `FileDescriptorSet`. What if we simplified it as:

```
This command compiles your proto files with "protoc", converts JSON input to binary
and converts the result from binary to JSON. All these steps take on the order of 
milliseconds. For example, the overhead for a file with four dependencies is about
30ms, so there is little overhead for CLI calls to gRPC.
```

Commenting here to get both your thoughts, not sure if this is any clearer.

peter-edge:
I'd like to merge this before doing https://github.com/uber/prototool/issues/16 - can we do what we'd like and get this in?

peter-edge:
Anything is fine. Ping @amckinney.

peter-edge:
Merging per email with @AlexAPB.

PR #153, opened 2018-07-17
Print single-line field options on the same line from the formatter

### LINKED ISSUE #108: Print single-line field options on the same line from the formatter
If you have a field:

```
int woot = 5 [ (bar.field_option) = true ];
```

Print that out instead of the current:

```
int woot = 5 [
  (bar.field_option) = true
];
```

peter-edge:
Fixes #108. See the diff on the format golden files below.

`POptions` effectively becomes `pOptions` (signaling it should not be called outside of `baseVisitor`), then the new `POptions` no longer takes `isFieldOption` bool, and fields/enum fields are special-cased in`PField` and `PEnumField`.

The code is not clean, but this is a larger symptom of how the code already is in `internal/format`, and of emicklei/proto. This is one of the smaller "transformations" from the existing code that we can make to fix #108. Let me know what you want to do with it!

### DIFF
--- a/internal/format/base_visitor.go
+++ b/internal/format/base_visitor.go
@@ -89,7 +89,7 @@ func (v *baseVisitor) pMessageOrEnumField(prefix string, fieldName string, field
 	}
 	if len(options) == 1 {
 		o := options[0]
-		if len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 {
+		if isSingleValueLiteral(o.Constant) {
 			if source := o.Constant.SourceRepresentation(); source != "" {
 				v.PWithInlineComment(inlineComment, prefix, fieldType, fieldName, " = ", fieldTag, " [", o.Name, ` = `, source, "];")
 				return
@@ -122,8 +122,7 @@ func (v *baseVisitor) pOptions(isFieldOption bool, options ...*proto.Option) {
 			}
 		}
 		v.PComment(o.Comment)
-		// TODO: this is a good example of the reasoning for https://github.com/uber/prototool/issues/1
-		if len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 {
+		if isSingleValueLiteral(o.Constant) {
 			// SourceRepresentation() returns an empty string if the literal is empty
 			// if empty, we do not want to print the key or empty value
 			if source := o.Constant.SourceRepresentation(); source != "" {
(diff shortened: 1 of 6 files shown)


internal/format/base_visitor.go
```diff
@@ -152,17 +182,16 @@ func (v *baseVisitor) pInnerLiteral(name string, literal proto.Literal, suffix s
...
 
-func (v *baseVisitor) PField(prefix string, t string, field *proto.Field) {
-	v.PComment(field.Comment)
-	if len(field.Options) == 0 {
-		v.PWithInlineComment(field.InlineComment, prefix, t, " ", field.Name, " = ", field.Sequence, ";")
+func (v *baseVisitor) PField(prefix string, fieldType string, field *proto.Field) {
+	v.pMessageOrEnumField(prefix, field.Name, fieldType, field.Sequence, field.Comment, field.InlineComment, field.Options...)
+}
+
+func (v *baseVisitor) PEnumField(element *proto.EnumField) {
```
amckinney: I'd prefer to handle the case of a `nil` `element.ValueOption` option in the `pMessageOrEnumField` func so that this can simply be:
```
v.pMessageOrEnumField("", element.Name, "", element.Integer, element.Comment, element.InlineComment, element.ValueOption)
```

But I realize that this change would also need to propagate to `pOptions`, and the change would complicate things a bit.

Eh, I'm fine with leaving this in as is.

internal/format/base_visitor.go
```diff
@@ -73,7 +73,37 @@ func (v *baseVisitor) PComment(comment *proto.Comment) {
...
+		fieldType = fieldType + " "
+	}
+	v.PComment(comment)
+	if len(options) == 0 {
+		v.PWithInlineComment(inlineComment, prefix, fieldType, fieldName, " = ", fieldTag, ";")
+		return
+	}
+	if len(options) == 1 {
+		o := options[0]
+		if len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 {
```
amckinney: Seeing as this is used in `pOptions` as well and has an issue associated with it, we might want to make a more explanatory helper function. I'm not sure what a good name for this would be:

```
func <someName>(o *proto.Option) bool {
  return len(o.Constant.Array) == 0 && len(o.Constant.OrderedMap) == 0 && o.Constant.SourceRepresentation() != ""
}
```

peter-edge: Done

peter-edge:
All comments addressed @amckinney

PR #155, opened 2018-07-18
Remove the protoc-url global flag

amckinney:
Putting this up for discussion -

Given that we intend all of our users to specify the desired `protoc_version` in their `prototool.yaml`, it might not be necessary to support the `protoc-url` override in our global flag set.

I'd rather remove this override for now if we don't currently have a need for it. I can see where it might be useful (fetching a `protoc` zip file from another location), but this might better introduced as a `prototool.yaml` setting, and not as a global flag.


amckinney:
Ok, makes sense. Is this a flag so that these cases don't need a `prototool.yaml`?

PR #157, opened 2018-07-25
Allow file urls for the --protoc-url flag

### LINKED ISSUE #144: Re-add download as a context-specific command
Right now, `download` uses whatever configuration is in your current directory, so for example if your current directory has no `prototool.yaml` file, and sub-directory `a/b` has `a/b/prototool.yaml` with `protoc_version: 3.4.0`, `prototool download` will download the default version of `protoc` (currently 3.5.1), not the version in the `prototool.yaml` file. We should make `download` and `clean` take `dirOrProtoFiles...` so `prototool download a/b` does what you expect.

peter-edge:
This allows urls that start with `file://` for `-protoc-url` to allow for local download points. I've tested this manually. Fixes #144.


amckinney:
Let's add a changelog entry!

Issue #142, opened 2018-07-10
Merge FileDescriptorSets returned from Compiler into single FileDescriptorSet


peter-edge:
Now that we have one build configuration via allowing only one `prototool.yaml` file, all imports should be from the same places, therefore we should be able to smartly merge the `FileDescriptorSets` returned from `internal/protoc.Compiler` into a single `FileDescriptorSet`. This involves making sure all `FileDescriptorProtos` with the same name are actually equal, perhaps by a byte comparison of the serialized `FileDescriptorProtos`. If we have one `FileDescriptorSet`, a lot of operations in dependent packages get much simpler (grpc, #131, etc).

PR #158, opened 2018-07-25
Merge file descriptor sets

amckinney:
Re: https://github.com/uber/prototool/issues/142

Now that we only support a single `prototool.yaml`, we can safely merge the collected file descriptor sets into a single set. We simply loop over the separate file descriptor sets, and verify that any duplicate files have the same byte representation. We uniquely identify each file via `*descriptor.FileDescriptorProto.GetName` and compare their bytes via `*descriptor.FileDescriptorProto.String`, which marshals the file's contents under the hood.

Seeing as this is an internal implementation detail, I didn't include a changelog entry. Let me know if you'd like me to include a note there.

### DIFF
--- a/internal/protoc/compiler.go
+++ b/internal/protoc/compiler.go
@@ -161,8 +161,40 @@ func (c *compiler) Compile(protoSet *file.ProtoSet) (*CompileResult, error) {
 			fileDescriptorSets = append(fileDescriptorSets, fileDescriptorSet)
 		}
 	}
+	// Given that we only allow one build configuration, we can merge all of
+	// file descriptor sets into a single set.
+	fds, err := mergeFileDescriptorSets(fileDescriptorSets)
+	if err != nil {
+		return nil, fmt.Errorf("failed to merge file descriptors: %v", err)
+	}
 	return &CompileResult{
-		FileDescriptorSets: fileDescriptorSets,
+		FileDescriptorSet: fds,
+	}, nil
+}
+
+// Merge the file descriptor sets into a single set. All of the import paths should
+// refer to the same proto files, so we verify by comparing the raw bytes
+// between the files that are referenced more than once.
+func mergeFileDescriptorSets(sets []*descriptor.FileDescriptorSet) (*descriptor.FileDescriptorSet, error) {
+	// Used to determine whether or not a *descriptor.FileDescriptorProto
+	// has already been added to the merged set. We use the file descriptor's
+	// Name attribute to uniquely represent every file descriptor.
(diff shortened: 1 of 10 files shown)


internal/protoc/compiler.go
```diff
@@ -161,8 +161,40 @@ func (c *compiler) Compile(protoSet *file.ProtoSet) (*CompileResult, error) {
...
+// refer to the same proto files, so we verify by comparing the raw bytes
+// between the files that are referenced more than once.
+func mergeFileDescriptorSets(sets []*descriptor.FileDescriptorSet) (*descriptor.FileDescriptorSet, error) {
+	// Used to determine whether or not a *descriptor.FileDescriptorProto
+	// has already been added to the merged set. We use the file descriptor's
+	// Name attribute to uniquely represent every file descriptor.
+	files := map[string]*descriptor.FileDescriptorProto{}
+	for _, s := range sets {
+		for _, f := range s.File {
+			name := f.GetName()
```
peter-edge: I’m not confident this is 100% safe, I need to think about it more. The safest would be to marshal each FileDescriptorProto and hash the byte slice (cast to a string), and check if there are duplicates

AlexAPB:
I still need to dive into this. However, even though we closed the discussions on multiple `prototool.yaml` files for 1.0. It may be worth it to keep the current implementation in-case we ever find a strong reason too support multiple config files.

Maybe this change would be better post-1.0, thoughts? @amckinney @peter-edge

internal/protoc/compiler.go
amckinney: Yea, that's actually what we're doing here. `GetName` uniquely identifies the file and we use the `String` method to marshal that `FileDescriptorProto` into bytes and compare them. For more, see the description above.

amckinney:
Re: @AlexAPB 

Yea that's a great point. I'm fine with deferring this until later - the minor clean up in the respective commands (e.g. `GRPC`) probably isn't worth it for now.

I'll close this for now. We can re-open it and take it forward at a later time if there's ever a need.

internal/protoc/compiler.go
peter-edge: Oh whoops, missed the string thing! So just for clarity, it isn't technically marshalling to Protobuf, it's doing `proto.CompactTextString()` on the value. Should be the same effect in theory, just different serialization format, but for safety, I would worry that `CompactTextString` does not cover everything (with absolutely no research on the subject right now), and I would have preffered to do `proto.Marshal()` - just for future reference!

PR #162, opened 2018-07-30
Simplify external configuration settings

amckinney:
This un-registers and renames the external configuration settings according to the following:

- **Remove no_default_excludes** (true by default)
- **Remove include_well_known_types** (true by default)
- **Remove gen.go_options.no_default_modifiers** (error if set)
- **Remove lint.group** (error if set)
- **Rename create.dir_to_base_package to create.dir_to_package**

Note that this PR is currently branched off of https://github.com/uber/prototool/pull/161. Many of the testdata file directories are updated there, so this PR can be more-easily merged in afterwards.


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Delete the ability to explicitly specify multiple files, and have the effect
   of one file being specified be the same as the former `--dir-mode`. See
   [#16](https://github.com/uber/prototool/issues/16) for more details.
+- Remove protoc_include_wkt setting. This is always set to true.
```
peter-edge: Stupid things:
- s/Remove/Delete/ to be consistent with above
- Put tickmarks around each option name ie `protoc_include_wkt`.

internal/cmd/testdata/lint/allgroup/prototool.yaml
```diff
@@ -1,3 +1,14 @@
 protoc_include_wkt: true
 lint:
-  group: all
```
peter-edge: Ah shoot, this is why I had it be a soft failure if `lint.group` is specified - this test tests if all linters are turned on. I'm not sure what to do here - we could have all these options be an error if not in devel mode? But then you have to propagate development mode all the way down, which I don't like...I don't know ugh.

I'm moderately confused as to how tests are passing if you remove this.

internal/settings/settings.go
```diff
@@ -290,6 +290,24 @@ type ExternalConfig struct {
 	} `json:"gen,omitempty" yaml:"gen,omitempty"`
 }
 
+// Validate returns an error if any of the restricted, yet previously supported,
+// ExternalConfig attributes are set. This includes:
+//   - NoDefaultExcludes
+//   - Lint.Group
+//   - Gen.GoOptions.NoDefaultModifiers
+func (e ExternalConfig) Validate() error {
+	if e.NoDefaultExcludes {
```
peter-edge: Actually, we're going to want to make this the new default, and make this true, or alternatively remove anything from the list of default excludes in https://github.com/uber/prototool/blob/dev/internal/settings/settings.go#L49

internal/cmd/testdata/lint/allgroup/prototool.yaml
amckinney: I fixed the test by adding all of the missing linters under `include_ids`. This way, we mimic the `all` setting because it's essentially: `Default + (All - Default) == All`

CHANGELOG.md
amckinney: Sure, sounds good!

internal/settings/settings.go
amckinney: Ah yea good point. Nice catch, I'll update.

amckinney:
@peter-edge All comments addressed.

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Delete the ability to explicitly specify multiple files, and have the effect
   of one file being specified be the same as the former `--dir-mode`. See
   [#16](https://github.com/uber/prototool/issues/16) for more details.
+- Delete `protoc_include_wkt` setting. This is always set to true.
+- Delete `no_default_excludes` setting. This is always set to true.
+- Delete `gen.go_options.no_default_modifiers` setting.
+- Delete `lint.group` setting.
+- Delete `create.dir_to_base_package` -> `create.dir_to_package`.
```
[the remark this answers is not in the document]
amckinney: Ah yea, that was a bad update on my end. I'll update before rebasing / merging into `dev`.

mickeyreiss:
Wanted to provide some feedback that this breaks our working v0.4.0 configuration. I understand the project is in pre-1.0, but it would be helpful if breaking changes in the external API along with migration notes, for example, in the readme or changelog. 😄

peter-edge:
All changes have been noted in the changelog

PR #164, opened 2018-08-01
Fix excludes prefix matching

amckinney:
Previously, the `excludes` setting had the following comment:
```
These are prefixes, not regexes, so path/to/a will ignore anything
beginning with $(dirname some/dir/prototool.yaml)/path/to/a
including for example $(dirname some/dir/prototool.yaml)/path/to/ab
```
This fixes the prefix behavior so that we can exclude `./path/to/a/` without excluding `./path/to/ab`.

Note that this also makes a small cleanup with respect to the `NoDefaultExcludes` setting. Many of the configuration settings are left in the implementation so that they can be easily introduced later (if need be). I think it's safe to remove this particular setting though.

Default exclusions can vary from language to language, there likely won't ever be a need such a setting. Additionally, re-introducing something like this would be simple even without a reference in the implementation. I don't expect any conflict in this regard, so I included that change here since it pertains to the `excludes` setting. Please let me know if you disagree!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Delete `gen.go_options.no_default_modifiers` setting.
 - Delete `lint.group` setting.
 - Rename `create.dir_to_base_package` -> `create.dir_to_package`.
+### Fixed
 - Fix `excludes` setting to correctly match file path prefixes.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Delete `gen.go_options.no_default_modifiers` setting.
 - Delete `lint.group` setting.
 - Rename `create.dir_to_base_package` -> `create.dir_to_package`.
+- Fix `excludes` setting to correctly match file path prefixes.
 
 
 ## [0.5.0] - 2018-07-26
--- a/internal/file/proto_set_provider.go
+++ b/internal/file/proto_set_provider.go
@@ -293,7 +293,7 @@ func getProtoFiles(filePaths []string) ([]*ProtoFile, error) {
 // isExcluded determines whether the given filePath should be excluded.
 // Note that all excludes are assumed to be cleaned absolute paths at
 // this point.
-// stopPath represnts the absolute path to the prototool configuration.
+// stopPath represents the absolute path to the prototool configuration.
 // This is used to determine when we should stop checking for excludes.
 func isExcluded(filePath, stopPath string, excludes map[string]struct{}) bool {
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Delete `gen.go_options.no_default_modifiers` setting.
 - Delete `lint.group` setting.
 - Rename `create.dir_to_base_package` -> `create.dir_to_package`.
+- Fix `excludes` setting to correctly match file path prefixes.
```
AlexAPB: under `### Fixed` header?

internal/settings/settings.go
```diff
@@ -259,7 +259,6 @@ type OutputPath struct {
 // It is meant to be set by a YAML or JSON config file, or flags.
 type ExternalConfig struct {
 	Excludes           []string `json:"excludes,omitempty" yaml:"excludes,omitempty"`
-	NoDefaultExcludes  bool     `json:"no_default_excludes,omitempty" yaml:"no_default_excludes,omitempty"`
```
AlexAPB: this should get a changelog entry

internal/file/proto_set_provider.go
```diff
@@ -298,3 +289,32 @@ func getProtoFiles(filePaths []string) ([]*ProtoFile, error) {
 	}
 	return protoFiles, nil
 }
+
+// isExcluded determines whether the given filePath should be excluded.
+// Note that all excludes are assumed to be cleaned absolute paths at
+// this point.
+// stopPath represnts the absolute path to the prototool configuration.
```
AlexAPB: nit: `represents`

internal/settings/config_provider.go
```diff
@@ -264,30 +264,22 @@ func externalConfigToConfig(e ExternalConfig, dirPath string) (Config, error) {
...
 	if err := yaml.Unmarshal(data, &s); err != nil {
 		return nil, err
 	}
-	return getExcludePrefixes(s.ExcludePaths, s.NoDefaultExcludePaths, dirPath)
+	return getExcludePrefixes(s.ExcludePaths, dirPath)
 }
 
-func getExcludePrefixes(excludes []string, noDefaultExcludes bool, dirPath string) ([]string, error) {
-	if !noDefaultExcludes {
-		excludes = append(DefaultExcludePrefixes, excludes...)
```
AlexAPB: Looks like we're removing `DefaultExcludePrefixes` entirely let's remove it's declaration from [here](https://sourcegraph.com/github.com/uber/prototool@50b7e7e874bde45db3da837f087cc7ba205434ec/-/blob/internal/settings/settings.go#L49:2).

I can't really see us re-introducing this, plus it's a minimal change if we want to add it back.

internal/settings/settings.go
peter-edge: It already has one https://github.com/uber/prototool/blob/dev/CHANGELOG.md

PR #165, opened 2018-08-02
Move plugin override to plugin path

amckinney:
This removes `gen.plugin_overrides` and adds the `gen.plugins.path` setting in its place. Note that this is functionally equivalent.

I verified this locally and continued to see the expected `--plugin=...` flag in the `--dry-run` output.


amckinney:
Seeing as this was a fairly lightweight change, I relied on some local testing (as mentioned in the description) along with exception-driven testing.

PR #169, opened 2018-08-03
Update Lint configuration

amckinney:
Linters are currently configured in the following way:

```
# Lint directives.
lint:
  # Linter * files to ignore.
  ignore_id_to_files:
    ...
  # The specific linters to use.
  ids:
    ...
  # Linters to include that are not in the default lint group.
  include_ids:
    ...
  # Linters to exclude from the default lint group.
  exclude_ids:
    ...
```

This proposes the following change:
```
# Lint directives.
lint:
  # Linter files to ignore.
  ignores:
    - id: <id>
      files:
        ...
    - id: <id>
      files:
        ...

# Linter rules.
  rules:
  # Determines whether or not to include the default set of linters.
  no_default: true
  # The specific linters to add.
  add:
    ...
  # The specific linters to remove.
  remove:
    ...
```


internal/lint/lint.go
```diff
@@ -171,58 +171,42 @@ func NewLinter(id string, purpose string, addCheck func(func(*text.Failure), str
 
 // GetLinters returns the Linters for the LintConfig.
 //
-// The config is expected to be valid, ie slices deduped, all upper-case,
-// and only either IDs or Group/IncludeIDs/ExcludeIDs, with no overlap between
-// IncludeIDs and ExcludeIDs.
+// The configuration is expected to be valid, deduplicated, and all upper-case.
+// IncludeIDs and ExcludeIDs MUST NOT have an intersection.
```
AlexAPB: optional: We could add a validation step to ensure that `IncludeIDs` and `ExcludeIDs` are mutually exclusive and return an error. This would prevent users from relying on undefined behaviour.

amckinney: We actually already have this in place. You can see it in the current master here: https://github.com/uber/prototool/blob/dev/internal/settings/config_provider.go#L260

PR #170, opened 2018-08-03
Replace `format --no-rewrite` with `format --fix`

amckinney:
This puts us in a better position with respect to backwards-compatibility / expected usability. By removing the default "rewrite package options" behavior from `format`, users will now opt-in to fixing their `.proto` files with respect to a variety of conditions (potentially introduced at a later time). By default, `format` will only concern itself with syntax, not semantics.

Currently, the `--fix` flag takes on the behavior of the previous `format` without any flags.


README.md
```diff
@@ -139,15 +139,10 @@ Lint your Protobuf files. The default rule set follows the Style Guide at [etc/s
 Format a Protobuf file and print the formatted file to stdout. There are flags to perform different actions:
 
 - `-d` Write a diff instead.
+- `-f` Fix the file according to the Style Guide.
 - `-l` Write a lint error in the form file:line:column:message if a file is unformatted.
 - `-w` Overwrite the existing file instead.
 
-By default, the values for `java_multiple_files`, `java_outer_classname`, and `java_package` are updated
```
peter-edge: I would keep this whole block as an explanation for what fix currently does - HTTP GET the contents of the README doesn’t have to produce backwards compatible output lol

PR #193, opened 2018-08-16
WIP/Demo Only: Protoc Alpine issue

peter-edge:
See #194.


peter-edge:
Update: if I install `libc6-compat`, it works. We've verified that `protoc` is not statically linked.

```
RUN \
  apk update && \
  apk add libc6-compat && \
  rm -rf /var/cache/apk/*
```

PR #195, opened 2018-08-18
Add output flag to version subcommand

mickeyreiss:
I was writing a makefile that wraps up prototool, and realized it would be useful if I could programmatically check the version of prototool. To do this, I implemented the `--json` flag on the `version` subcommand:

```sh
# Default: Tab Format (no change here, backwards compatible)
$ prototool version
Version:                 1.0.0-dev
Default protoc version:  3.6.1
Go version:              go1.10.3
OS/Arch:                 darwin/amd64

# JSON
$ prototool version --json
{
  "version": "1.0.0-dev",
  "default_protoc_version": "3.6.1",
  "go_version": "go1.10.3",
  "git_commit": "",
  "built_timestamp": "",
  "GOOS": "darwin",
  "GOARCH": "amd64"
}
```

### DIFF
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -43,6 +43,7 @@ type flags struct {
 	listLinters    bool
 	lintMode       bool
 	method         string
+	outputFormat   string
 	overwrite      bool
 	pkg            string
 	printFields    string
@@ -117,6 +118,10 @@ func (f *flags) bindListLinters(flagSet *pflag.FlagSet) {
 
 func (f *flags) bindMethod(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.method, "method", "", "The GRPC method to call in the form package.Service/Method. This is required.")
+}
+
+func (f *flags) bindOutput(flagSet *pflag.FlagSet) {
+	flagSet.StringVarP(&f.outputFormat, "output", "", "", "Output format: JSON or Tabs")
 }
 
 func (f *flags) bindOverwrite(flagSet *pflag.FlagSet) {
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -43,7 +43,6 @@ type flags struct {
 	listLinters    bool
 	lintMode       bool
 	method         string
-	outputFormat   string
 	overwrite      bool
 	pkg            string
 	printFields    string
@@ -118,10 +117,6 @@ func (f *flags) bindListLinters(flagSet *pflag.FlagSet) {
 
 func (f *flags) bindMethod(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.method, "method", "", "The GRPC method to call in the form package.Service/Method. This is required.")
-}
-
-func (f *flags) bindOutput(flagSet *pflag.FlagSet) {
-	flagSet.StringVarP(&f.outputFormat, "output", "", "", "Output format: JSON or Tabs")
 }
 
 func (f *flags) bindOverwrite(flagSet *pflag.FlagSet) {
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -47,7 +47,7 @@ func (e *ExitError) Error() string {
 type Runner interface {
 	Init(args []string, uncomment bool) error
 	Create(args []string, pkg string) error
-	Version(output string) error
+	Version() error
 	Download() error
 	Clean() error
 	Files(args []string) error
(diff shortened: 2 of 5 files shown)


internal/cmd/flags.go
```diff
@@ -114,6 +115,10 @@ func (f *flags) bindMethod(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.method, "method", "", "The GRPC method to call in the form package.Service/Method. This is required.")
 }
 
+func (f *flags) bindOutput(flagSet *pflag.FlagSet) {
+	flagSet.StringVarP(&f.outputFormat, "output", "", "", "Output format: JSON or Tabs")
```
peter-edge: If we're doing to do this, this should be `--json` or similar, or explicitly say what the values are here.

mickeyreiss: Sure, let me know what API(s) you'd prefer, and I'd be happy to make the change. I don't know, for example, whether prototool intends to implement output formats as a general feature in the future.

amckinney: Hey @mickeyreiss, thanks for this! Please use a `--json` flag similar to https://github.com/uber/prototool/pull/212. 

Otherwise, this approach looks good!

mickeyreiss:
@peter-edge I've pushed a quick set of changes to merge `uber:dev` and use the same `--json` flag as the one introduced in #212.

internal/cmd/flags.go
mickeyreiss: This should be addressed by a5b04fb and 9d809df.

internal/exec/runner.go
```diff
@@ -90,31 +90,58 @@ func newRunner(workDirPath string, input io.Reader, output io.Writer, options ..
...
-	if vars.GitCommit != "" {
-		if _, err := fmt.Fprintf(tabWriter, "Git commit:\t%s\n", vars.GitCommit); err != nil {
+func (r *runner) Version(json bool) error {
+	out := &struct {
+		Version              string `json:"version"`
+		DefaultProtocVersion string `json:"default_protoc_version"`
+		GoVersion            string `json:"go_version"`
+		GitCommit            string `json:"git_commit"`
+		BuiltTimestamp       string `json:"built_timestamp"`
+		GOOS                 string `json:"GOOS"`
```
peter-edge: Lowercase json tags

peter-edge: Also add omitempty to all tags

internal/exec/exec.go
```diff
@@ -47,7 +47,7 @@ func (e *ExitError) Error() string {
 type Runner interface {
 	Init(args []string, uncomment bool) error
 	Create(args []string, pkg string) error
-	Version() error
+	Version(output string) error
```
peter-edge: json bool

internal/exec/runner.go
```diff
@@ -90,31 +90,58 @@ func newRunner(workDirPath string, input io.Reader, output io.Writer, options ..
...
-	if _, err := fmt.Fprintf(tabWriter, "Default protoc version:\t%s\n", vars.DefaultProtocVersion); err != nil {
-		return err
-	}
-	if _, err := fmt.Fprintf(tabWriter, "Go version:\t%s\n", runtime.Version()); err != nil {
-		return err
-	}
-	if vars.GitCommit != "" {
-		if _, err := fmt.Fprintf(tabWriter, "Git commit:\t%s\n", vars.GitCommit); err != nil {
+func (r *runner) Version(json bool) error {
+	out := &struct {
```
peter-edge: No need to be a pointer, just do struct

internal/exec/runner.go
```diff
@@ -90,31 +90,58 @@ func newRunner(workDirPath string, input io.Reader, output io.Writer, options ..
...
+		GoVersion:            runtime.Version(),
+		GitCommit:            vars.GitCommit,
+		BuiltTimestamp:       vars.BuiltTimestamp,
+		GOOS:                 runtime.GOOS,
+		GOARCH:               runtime.GOARCH,
+	}
+
+	if json {
+		enc := json.NewEncoder(r.output)
+		enc.SetIndent("", "  ")
```
peter-edge: Why not just use MarshalIndent and runner.println?

peter-edge: @amckinney can’t remember what you did, but we should be consistent (and I think MarshalIndent would be the best across all json calls)

internal/exec/runner.go
mickeyreiss: Done in 7668c8e6169c8b758e77784caedeff222c45fca1

internal/exec/runner.go
mickeyreiss: I did this because it seemed like the simpler way to marshal to an `io.Writer` (`runner.output`).

peter-edge:
I edited the PR for @mickeyreiss - he's been extremely patient with this, and we appreciate it. Merging now.

PR #198, opened 2018-08-21
Add go1.11rc1 to .travis.yml

peter-edge:


peter-edge:
We might want to. This was to check to make sure everything still is building properly, but up to you and @amckinney whether to merge this.

Issue #204, opened 2018-08-23
Automated GitHub Releases


abhinav:
The current release process involves building the binaries on a laptop and
uploading them to GitHub manually. We can automate this: Build the binaries in
CI and upload them directly to GitHub from there when a tag in the form
`vX.Y.Z` is pushed.

We do this already in yab. See yab's [.travis.yml][1] and [release.sh][2] for
inspiration. We should be able to continue using the releasegen.sh script and
adapt yab's extract_changelog script to prototool's changelog format.

There's some credentials-related setup involved. We should use one of our bot
credentials to cut the releases.

  [1]: https://github.com/yarpc/yab/blob/dev/.travis.yml
  [2]: https://github.com/yarpc/yab/blob/dev/scripts/release.sh

peter-edge:
Would be nice - not sure it saves too much time or adds consistency (see below), but automation is generally good.

Note the build happens within a Docker container so that when there are panics, the location shows as `/go/src/...`. That should stay the same - Travis CI has Docker, so this shouldn't be too hard. Really the release process should remove the `vendor` directory and re-install as well, so that there's no chance of bad dependencies happening.

peter-edge:
Closing this for now - working towards Inbox Zero, I don't think we're going to get to this, we can revisit if this becomes an issue later.

PR #205, opened 2018-08-24
Add script to generate assets for Homebrew and update README.md installation instructions

### LINKED ISSUE #191: Add instructions to README.md about installing from release tarballs
The release tarballs also include bash completion, zsh completion, and man pages, and are meant to be untarred directly into i.e. `/usr/local`.

### LINKED ISSUE #203: Homebrew Formula and Bottle
We should make prototool easily installable by providing a Homebrew Formula
and Bottle. We can create our own Tap or place it in the default Formulae
collection (homebrew-core). An advantage of our own Tap is that the complete
release can happen in one go versus waiting on homebrew-core to accept the
change.

Ideally, updating the Formula and the Bottle is automated as part of some
release script.

This isn't really high priority and we can defer it until there's more demand
for it.

peter-edge:
This fixes #203, and also inadvertently fixes #191. This also updates the zsh location for `/usr/local` to the correct location, and calls `rm -rf vendor` and then `glide install` in `etc/bin/releasegen.sh` as I mentioned we should do in #204.

This adds helper functionality to build Prototool for Homebrew. In #203, there's a question about whether to do this in our own Tap versus homebrew-core - I personally vote for homebrew-core, as once the initial formula is added, the maintainers are *extremely* quick to update to new versions, and it reduces the complexity for external developers. We've gotten some great feedback, including discovery of some nasty bugs, from external developers, so anything we can do to increase that pipeline, especially now that this is v1.0, would be valuable. If there's a separate tap, external developers will instinctively be less likely to install (at least that would be my instinct, for whatever that's worth).

For the same reason, I'm in favor of doing this sooner than later - customer demand is sort of a chicken and egg problem, if we make it easier for customers to install Prototool, we'll have more demand. @AlexAPB @amckinney @abhinav thoughts?

This adds `etc/bin/brewgen.sh`, which is effectively a customized `etc/bin/releasegen.sh`. We could merge the two scripts, but doing so (and then having options in each script) ends up being just as complicated, if not more bug prone.

This also adds `make brewgen`, which is what the formula calls. We could just call the script directly in the Homebrew formula and remove the `automake` dependency, but we later might want to move more functionality into the Makefile for the build, so this makes it easier to do so.

The corresponding formula can be seen at https://github.com/peter-edge/homebrew-core/compare/master...peter-edge:prototool-1.0.0?expand=1 - note that `--HEAD` will not work right now, as this PR would need to be merged first. I've tested this locally, and it's pretty cool.

```
$ brew info prototool
prototool: stable v1.0.0, HEAD
Your Swiss Army Knife for Protocol Buffers
https://github.com/uber/prototool
/usr/local/Cellar/prototool/v1.0.0 (17 files, 16.6MB) *
  Built from source on 2018-08-24 at 17:26:56
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/prototool.rb
==> Dependencies
Build: automake ✔, glide ✔, go ✔
==> Options
--HEAD
	Install HEAD version
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh functions have been installed to:
  /usr/local/share/zsh/site-functions
```

This formula will also need to be updated - assumedly, if we merge this, we'll need to do a `v1.1.0` for the changes here, and then the Formula will be updated to have this version and sha256, as opposed to pointing to this branch.

Homebrew is really easy to use, so this results in the bash completion, zsh completion, and man pages being installed as well. Future updates to homebrew-core will be extremely easy since all the build functionality is wrapped in the `etc/bin/brewgen.sh` script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /.glide
 /.idea
+/brew
 /example/gen/proto/java
 /cover
 /cover.html
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,10 @@ releasegen: internalgen
 		$(DOCKER_IMAGE) \
 		bash -x etc/bin/releasegen.sh
 
+.PHONY: brewgen
+brewgen:
+	sh etc/bin/brewgen.sh
+
 .PHONY: releaseinstall
 releaseinstall: releasegen releaseclean
 	tar -C /usr/local --strip-components 1 -xzf release/prototool-$(shell uname -s)-$(shell uname -m).tar.gz
--- a/README.md
+++ b/README.md
@@ -37,7 +37,22 @@ Prototool accomplishes this by downloading and calling protoc on the fly for you
 
 ## Installation
 
-Install Prototool from GitHub Releases.
+Prototool can be installed on Mac OS X via [Homebrew](https://brew.sh/).
+
+```
+brew install prototool
+```
+
+This installs the `prototool` binary, along with bash completion, zsh completion, and man pages.
+You can also install all of the assets on Linux or without Homebrew from GitHub Releases.
+
+```
+curl -sSL https://github.com/uber/prototool/releases/download/v1.0.0/prototool-$(uname -s)-$(uname -m).tar.gz | \
+  tar -C /usr/local --strip-components 1 -xz`
+```
+
+If you do not want to install bash completion, zsh completion, or man mages, you can install just the
+`prototool` binary from GitHub Releases as well.
 
 ```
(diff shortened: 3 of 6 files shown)


peter-edge:
Good point, will add that to the RELEASE.md.

One other note - [Bottles](https://docs.brew.sh/Bottles.html) should be automatically built by Homebrew, so most people will not need `automake, glide, go` to install this once they automatically update the formula, and will not need to build from source.

PR #207, opened 2018-08-26
Removed backquote from  installation cmd of without homebrew

nametake:


nametake:
OK.

I would like you to correct it instead of me.

Thank you.

PR #212, opened 2018-08-28
Add JSON flag for JSON-encoded compile and lint failures

amckinney:
We originally supported a `--harbormaster` flag to format our `protoc` and `lint` failures in a JSON format. This tied us to the Harbormaster API spec as detailed [here](https://secure.phabricator.com/conduit/method/harbormaster.sendmessage/). 

This proposes a more general solution that formats these failures in terms of Prototool's vocabulary (i.e. `lintID` instead of `code`, and removing `severity` and `name` altogether).

The implementation is largely inspired by https://github.com/uber/prototool/pull/53, but we can now include the `json` tags on the `text.Failure` directly because it includes all of the fields we seek to model.

Example output for `compile` and `lint` is shown below:
```
$ prototool compile --json
{"filename":"idl/foo/foo.proto","line":13,"column":14,"message":"\"Foo\" is not defined."}

$ prototool lint --json
{"filename":"idl/foo/foo.proto","lint_id":"FILE_OPTIONS_REQUIRE_JAVA_MULTIPLE_FILES","message":"File option \"java_multiple_files\" is required."}
```

Note that this will be potentially relevant for other commands that could leverage JSON output, such as in https://github.com/uber/prototool/pull/195.

### DIFF
--- a/internal/text/text.go
+++ b/internal/text/text.go
@@ -111,7 +111,7 @@ type Failure struct {
 	Filename string `json:"filename,omitempty"`
 	Line     int    `json:"line,omitempty"`
 	Column   int    `json:"column,omitempty"`
-	LintID   string `json:"lintID,omitempty"`
+	LintID   string `json:"lint_id,omitempty"`
 	Message  string `json:"message,omitempty"`
 }
 
--- a/internal/text/text.go
+++ b/internal/text/text.go
@@ -111,7 +111,7 @@ type Failure struct {
 	Filename string
 	Line     int
 	Column   int
-	ID       string
+	LintID   string
 	Message  string
 }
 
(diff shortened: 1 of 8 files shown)


internal/text/text.go
```diff
@@ -108,11 +108,11 @@ func ParseColonSeparatedFailureFields(s string) ([]FailureField, error) {
...
 type Failure struct {
-	Filename string
-	Line     int
-	Column   int
-	ID       string
-	Message  string
+	Filename string `json:"filename,omitempty"`
+	Line     int    `json:"line,omitempty"`
+	Column   int    `json:"column,omitempty"`
+	LintID   string `json:"lintID,omitempty"`
```
peter-edge: this should probably be `lint_id`, to be consistent with our json/yaml in the rest of Prottoool

peter-edge:
Follow-up with `grpc` command, perhaps taking parts of https://github.com/uber/prototool/pull/104? :-)

internal/text/text.go
amckinney: I went with camelCase here after I found a note on it from Google's style guide: https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format

I'm fine with using snake_case in this case though since it's more consistent with the rest of Prototool.

amckinney:
Yeah we definitely want to follow-up on the `grpc` flag, as well. That's next! :)

PR #216, opened 2018-08-31
Add support for transport credentials

Willyham:
This adds the ability to configure transport credentials from a file for `prototool grpc`. I'd appreciate some help testing this as I wasn't able to run it against one of our services due to (what I believe is) an unrelated error. However, running without the `--certificate` flag gives a connection error, and running with it just gives me one related to the method, so I believe that it is working correctly.

### DIFF
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -25,30 +25,31 @@ import (
 )
 
 type flags struct {
-	address        string
-	cachePath      string
-	callTimeout    string
-	connectTimeout string
-	data           string
-	debug          bool
-	diffMode       bool
-	disableFormat  bool
-	disableLint    bool
-	dryRun         bool
-	fix            bool
-	headers        []string
-	keepaliveTime  string
-	json           bool
-	listAllLinters bool
-	listLinters    bool
-	lintMode       bool
-	method         string
-	overwrite      bool
-	pkg            string
-	printFields    string
-	protocURL      string
-	stdin          bool
-	uncomment      bool
+	address         string
+	cachePath       string
+	callTimeout     string
+	certificateFile string
+	connectTimeout  string
+	data            string
+	debug           bool
+	diffMode        bool
+	disableFormat   bool
+	disableLint     bool
+	dryRun          bool
+	fix             bool
+	headers         []string
+	keepaliveTime   string
+	json            bool
+	listAllLinters  bool
+	listLinters     bool
+	lintMode        bool
+	method          string
+	overwrite       bool
+	pkg             string
+	printFields     string
+	protocURL       string
+	stdin           bool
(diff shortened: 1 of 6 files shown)


amckinney:
Thanks for the contribution! We'll give this a closer look with respect to testing and get back to you.

internal/cmd/flags.go
```diff
@@ -63,6 +64,10 @@ func (f *flags) bindCallTimeout(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.callTimeout, "call-timeout", "60s", "The maximum time to for all calls to be completed.")
 }
 
+func (f *flags) bindCertificateFile(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.certificateFile, "certificate", "", "SSL Certificate to make requests with.")
```
peter-edge: Flag should either be named `--certificate-file` or variable should be named `certificate`.

internal/exec/runner.go
```diff
@@ -573,6 +575,7 @@ func (r *runner) GRPC(args, headers []string, address, method, data, callTimeout
 	var parsedCallTimeout time.Duration
 	var parsedConnectTimeout time.Duration
 	var parsedKeepaliveTime time.Duration
+	var creds credentials.TransportCredentials
```
peter-edge: `transportCredentials`, to keep the same style as the rest of the repository.

internal/cmd/flags.go
amckinney: We could be even more specific and name this `ssl-certificate` & `sslCertificate`.

internal/grpc/grpc.go
```diff
@@ -25,6 +25,8 @@ import (
 	"io"
 	"time"
 
+	"google.golang.org/grpc/credentials"
+
```
peter-edge: no newline

internal/cmd/flags.go
peter-edge: I personally vote for `--ssl-certificate-path` to continue with this `ITEM-NOUN` naming scheme. @amckinney thoughts?

internal/exec/runner.go
```diff
@@ -37,6 +37,8 @@ import (
 	"text/tabwriter"
 	"time"
 
+	"google.golang.org/grpc/credentials"
+
```
AlexAPB: nit: remove empty space

internal/cmd/flags.go
AlexAPB: either are fine by me, let's prefix with TLS instead of SSL, however

AlexAPB:
Closing due to inactivity, feel free to re-open if you get a chance.

We don't have the bandwidth to pull this over the line ourselves.

Issue #192, opened 2018-08-15
Move from glide to vgo and move internal/cmd/prototool back to cmd/prototool


peter-edge:
Something @amckinney and I discussed - this would allow versioned `go get`. The move from `internal/cmd/prototool` back to `cmd/prototool` would represent that we want this to be go-gettable (although technically both are). This would be backwards-compatible after v1.0 as we don't represent anything inside `internal` to be backwards-compatible.

PR #217, opened 2018-08-31
Promote cmd out of internal

amckinney:
In response to https://github.com/uber/prototool/issues/192, this moves the `./internal/cmd` package to `./cmd` so that it more clearly expresses the project's usability with `go get` when we introduce a `go.mod` file.

Note that this also allows projects to effectively version control `prototool`. 
For example, in dependency management systems like `glide`, we can add the following lines to the `glide.yaml`:
```
- package: github.com/uber/prototool
  version: v1.2.0
```

However, `glide` will not pick up the transitive dependencies for `prototool` by default unless the package is actually depended on. Now that `cmd` is out of `internal/`, we can add a simple import like so:
```
import _ "github.com/uber/prototool/cmd"
```


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+- Move `internal/cmd` to `cmd/` so that it can be depended on.
```
AlexAPB: `### Changed`

cmd/testdata/create/one/a/b/bar/baz.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package bat;
+
+option go_package = "batpb";
+option java_multiple_files = true;
+option java_outer_classname = "BazProto";
+option java_package = "com.bat";
```
AlexAPB: missing new lines at the end of these proto files

AlexAPB: Second thought, why are we introducing these files?

CHANGELOG.md
amckinney: Good catch - using `Added` since this is more additive from a user perspective. `Changed` is a change in behavior and sometimes implies a breaking change.

cmd/testdata/create/one/a/b/bar/baz.proto
amckinney: Oof, I forgot to update the `.gitignore` in my project-wide find/replace. Fixed 👍

amckinney:
@peter-edge Sorry, the description wasn't clear - you're not missing anything. We still can't use `go get` effectively until we have the `vgo` integration. The description would have  been better phrased as:
```
In response to #192, this moves the ./internal/cmd package to ./cmd so that it more clearly expresses the project's usability with go get when we introduce a go.mod file.
```

I was attempting to solve the version control issue (i.e. the `glide` example) and saw this as "killing two birds with one stone" since it would perform part of what is required in #192. Seeing as this doesn't _really_ change the behavior of `go get` at all, I thought it would better belong in a separate change. 

What are your thoughts? Next steps would be to include a `go.mod` file (potentially in a separate PR) and move the `gen*` commands back into `internal` packages (in this PR).

peter-edge:
My general thoughts:

- The only thing that should be made external eventually is `cmd/prototool`, not `cmd` as a whole (unless depending on `cmd` as a library is needed, but I think you'd want to expose individual parts, see #219), and for sure not the helper release commands :-)
- If you're going to make it external, you should do it only once you have the `go.mod` file in place, so that `go get` works as appropriate with Go 1.11 (and update the documentation to reflect that `go get` only works with Go 1.11+ as well, and only if `GO111MODULE=on` is set I believe). If you do this change as-is, it is an indicator that `go get` works right now, which would not be the case as the PR is currently written. That will mean more filed issues/installation confusion for end-users IMO.

peter-edge:
Also note I started doing the conversion in https://github.com/uber/prototool/pull/197, however there's a bunch of issues to doing this properly, most notably that you will want a build tag for testing dependencies, and especially for tool dependencies. See https://github.com/golang/go/issues/24661 to get started.

amckinney:
Sweet, that sounds reasonable. Thanks for the links! Let's give this a look and make sure everything is in place before we move this forward. I'm happy to defer / close this for now and re-open when it is relevant.

PR #220, opened 2018-09-05
Add installation support via a Homebrew Tap

peter-edge:
**Update** this does not work, the repository has to be prefixed with `homebrew-`. We're looking into other solutions, and keeping the homebrew-core PR open in the hope it gets merged.

Re: #203, this changes the Homebrew strategy from using homebrew-core to having Prototool be it's own Homebrew Tap https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md. https://github.com/Homebrew/homebrew-core/pull/31440 has been open for 12 days without it being merged, which concerns me in terms of future version updates. This allows us to have Prototool be a Homebrew Tap that can be updated when a new version is released.

Installation is just as simple:

```bash
brew install uber/prototool/prototool
```

Homebrew will get the Tap from the default branch (dev), so when we do a release, we compute the checksum of the newly released tarball and push the change to dev. The instructions in RELEASE.md are updated as well. Note there is no need to do a new release after this - the dev branch is where the Tap is read from, and the instructions in RELEASE.md update dev after the release tag and master have been pushed, which has to happen anyways as you need to compute the checksum of the released tarball for the Tap.

The downside to this is that we won't get pre-built binaries (which are automatically created via homebrew-core), so users will be building from source indefinitely as of now. However, we could make our own Bottles in the future https://docs.brew.sh/Bottles.html. Additionally, if we wanted to keep on pursuing https://github.com/Homebrew/homebrew-core/pull/31440 and are able to get versions merged more quickly, homebrew-core is prioritized over separate Taps, so if we later changed the advice to `brew install prototool`, this would take precedence over `uber/prototool/prototool` https://github.com/Homebrew/brew/blob/master/docs/Taps.md, meaning both could live next to each other simultaneously.

After this PR is merged, we need to quickly test to make sure this works as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+### Added
+- Add installation support via a Homebrew Tap.
 
 
 ## [1.2.0] - 2018-08-29
--- a/HomebrewFormula/prototool.rb
+++ b/HomebrewFormula/prototool.rb
@@ -1,0 +1,30 @@
+class Prototool < Formula
+  desc "Your Swiss Army Knife for Protocol Buffers"
+  homepage "https://github.com/uber/prototool"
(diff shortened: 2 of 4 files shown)


peter-edge:
Closing as the homebrew-core PR was merged.

PR #223, opened 2018-09-07
Accept prototool.json configuration files in addition to prototool.yaml configuration files

peter-edge:
This updates `internal/settings` to allow both `prototool.yaml` and `prototool.json` configuration files. The actual logic changes are minimal, and there is one cleanup, namely an anonymous struct is no longer used for exclude paths - this was a bit flaky as this struct would not get updated, instead we use the same logic for unmarshalling of configuration files regardless now.

The diff looks large, but the majority of it is just moving of files, and changing a few files to `prototool.json` instead of `prototool.yaml`. In the tests in `internal/file`, I had to separate `testdata` into `testdata/valid/base` and `testdata/invalid`, the latter of which just contains two files, `prototool.yaml` and `prototool.json`, to verify that an error is returned if we have multiple configuration files in one directory.


internal/settings/config_provider.go
```diff
@@ -83,41 +88,91 @@ func (c *configProvider) GetExcludePrefixesForDir(dirPath string) ([]string, err
...
 // given directory, and going up a directory until hitting root.
 //
 // The directory must be an absolute path.
 //
 // If no such file is found, "" is returned.
+// If multiple files named by one of the ConfigFilenames are found in the same
+// directory, error is returned.
 // Also returns all the directories this Config applies to.
-func getFilePathForDir(dirPath string) (string, []string) {
+func getFilePathForDir(dirPath string) (string, []string, error) {
```
amckinney: Do we need to return the `[]string` here? Now that we've updated the signature, I see that it's only been updated in one place and the value is ignored, i.e.
```
filePath, _, err := getFilePathForDir(dirPath)
```

Should we update the signature to:
```
func getFilePathForDir(dirPath string) (string, error)
```

internal/settings/config_provider.go
```diff
@@ -256,21 +311,18 @@ func externalConfigToConfig(e ExternalConfig, dirPath string) (Config, error) {
...
-	if err := yaml.Unmarshal(data, &s); err != nil {
+	if filePath == "" {
+		return []string{}, nil
+	}
+	externalConfig, err := getExternalConfig(filePath)
+	if err != nil {
 		return nil, err
 	}
-	return getExcludePrefixes(s.ExcludePaths, dirPath)
+	return getExcludePrefixes(externalConfig.Excludes, dirPath)
```
amckinney: Nice cleanup :+1:

internal/settings/settings.go
```diff
@@ -45,6 +45,12 @@ const (
 )
 
 var (
+	// ConfigFilenames are all possible config filenames.
+	ConfigFilenames = []string{
+		DefaultConfigFilename,
```
amckinney: It looks like we're using this for `prototool config init`. I imagine that we'll need to support a `prototool config init --json` now that we support `prototool.json`. 

Not actionable in this change, but worth noting for now.

peter-edge: I almost wanted to not document the `prototool.json` stuff as much for now, but thought that was probably not a good idea. I think that we should say we prefer `prototool.yaml` if you're generating a file, and `prototool.json` is sort of the "secondary" format, so if you use `prototool config init`, it only does YAML - at least I think this is fair for now.

Tldr: We're in agreement, maybe add later, heh.

internal/settings/config_provider.go
peter-edge: Doing this actually results in a bunch more cleanups, nice catch. Done.

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+- Accept `prototool.json` files for configuation in addition to
```
AlexAPB: `### Added` header

internal/settings/config_provider.go
```diff
@@ -83,41 +83,88 @@ func (c *configProvider) GetExcludePrefixesForDir(dirPath string) ([]string, err
...
+func getExternalConfig(filePath string) (ExternalConfig, error) {
+	data, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return ExternalConfig{}, err
+	}
 	externalConfig := ExternalConfig{}
-	if err := yaml.UnmarshalStrict(data, &externalConfig); err != nil {
-		return Config{}, err
+	switch filepath.Ext(filePath) {
+	case ".json":
```
AlexAPB: Formatting nit to reduce duplication: let's lift the `len(data)` checks above the `switch` and have one `return` after the switch statement (leaving the `default` case as is).

internal/settings/config_provider.go
```diff
@@ -83,41 +83,88 @@ func (c *configProvider) GetExcludePrefixesForDir(dirPath string) ([]string, err
...
+			return "", nil
 		}
 		dirPath = filepath.Dir(dirPath)
 	}
 }
 
+// getSingleFilePathForDir gets the file named by one of the ConfigFilenames in the
+// given directory. Having multiple such files results in an error being returned. If no file is
+// found, this returns "".
+func getSingleFilePathForDir(dirPath string) (string, error) {
```
AlexAPB: nit: can we just call this `getConfigPath` or similar? Using `single` seems a bit odd since we will only ever expect to have one file path for a configuration.

AlexAPB:
Also, agreed with you @amckinney this should be the last configuration file type we introduce.

internal/settings/config_provider.go
peter-edge: We can go through and refactor this ,but this would allow an empty file i.e. `prototool.toml` to produce no error with the current logic. If there's a cleanup refactor on this, let's do it in another PR, but I think the logic should stay as is.

CHANGELOG.md
peter-edge: Done

internal/settings/config_provider.go
peter-edge: This is matching the above `getFilePathForDir`, which performs the logic to go up to root for the given directory path. I think having `getFilePathForDir` and `getConfigPath` confuses the purpose of these two functions.

peter-edge:
Addressed PR comments - going with the approval and since we have a bunch outstanding, if there's disagreement on the nits feel free to put a cleanup PR up.

internal/settings/config_provider.go
AlexAPB: Should we really restrict people from having a `prototool.foo` in some directory? I _think_ this makes sense here to actually silently fail. 

It may actually be beneficial to allow a file like `prototool.sh` in a repo for running common commands. Now that I'm thinking about it, this may actually break some current users. Thoughts?

cc @amckinney

peter-edge: Good point - I don’t think this actually breaks anyone (the above code), but decent point on not switching until sure.

peter-edge: Moved the `len(data) == 0` check to the top, all comments should be addressed now

PR #224, opened 2018-09-07
Add --config-data flag

peter-edge:
This adds the `--config-data` flag.

Having played with this, I actually do wish we added a `--config-dir` or `--config-dir-path` flag as well, If you play with it locally you might see what I mean, but I understand the reasoning not to add it.

A nice representation of having a JSON file and then `cat`'ing it on the command line would be to cd into `internal/cmd/testdata/format/proto3` and then doing:

```
$ prototool lint
foo/foo.proto:54:3:C-Style comments are not allowed.
foo/foo.proto:112:3:Enum aliases are not allowed.

$ prototool lint --config-data "$(cat prototool.json)"
foo/foo.proto:54:3:C-Style comments are not allowed.
foo/foo.proto:112:3:Enum aliases are not allowed.
```

Note if you have bash completion set up on your computer and installed with `brew install prototool`, you need to do `brew uninstall prototool` for `prototool` to be picked up on your `$PATH` from `make install`, as bash completion will still make it so that the Homebrew `prototool` binary is used, regardless of the order of the directories in your `$PATH`.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 - Accept `prototool.json` files for configuation in addition to
   `prototool.yaml` files.
+- Add `--config-data` flag.
```
amckinney: Let's add more context here, such as how this can be used and/or is useful.

internal/cmd/flags.go
```diff
@@ -63,6 +64,10 @@ func (f *flags) bindCallTimeout(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.callTimeout, "call-timeout", "60s", "The maximum time to for all calls to be completed.")
 }
 
+func (f *flags) bindConfigData(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.configData, "config-data", "", "The configuration data to use instead of reading prototool.yaml or prototool.json files.\nThis will act as if there is a configuration file with the given data in the current directory, and no other configuration files recursively.\nThis is an advanced feature and is not recommended to be generally used.")
```
amckinney: I think we should shorten this description. The first sentence will likely suffice:
```
The configuration data to use instead of reading prototool.yaml or prototool.json files.
```

internal/cmd/templates.go
```diff
@@ -45,6 +45,7 @@ var (
 			return runner.All(args, flags.disableFormat, flags.disableLint, flags.fix)
 		},
 		BindFlags: func(flagSet *pflag.FlagSet, flags *flags) {
+			flags.bindConfigData(flagSet)
```
amckinney: Why not just make this a global flag, similar to `debug`? I realize that some commands might not need a configuration, such as `version` or `config init`, but the same could be said about `debug`, so adding this to the global set seems fine to me. What are your thoughts?

internal/file/proto_set_provider.go
```diff
@@ -33,6 +33,8 @@ import (
 
 type protoSetProvider struct {
 	logger         *zap.Logger
+	configDirPath  string
```
amckinney: We should be able to drop this field (see below).

internal/file/proto_set_provider.go
```diff
@@ -90,24 +92,30 @@ func (c *protoSetProvider) GetForFiles(workDirPath string, filePaths ...string)
...
 	if err != nil {
 		return nil, err
 	}
-	// we need everything for generation, not just the files in the given directory
-	// so we go back to the config file if it is shallower
-	// display path will be unaffected as this is based on workDirPath
-	configDirPath := absDirPath
-	if configFilePath != "" {
-		configDirPath = filepath.Dir(configFilePath)
+	configDirPath := c.configDirPath
```
amckinney: Let's remove `protoSetProvider.configDirPath` and just use the `workDirPath` instead. They should always be the same.

```
configDirPath := absDirPath
if c.configData {
  // The users has specified configuration via the command line. Set the configuration
  // directory to the current working directory.
  configDirPath = workDirPath
} else {
  configFilePath, err := c.configProvider.GetFilePathForDir(absDirPath)
  if err != nil {
    return nil, err
  }
  // We need everything for generation, not just the files in the given directory.
  // Traverse back to the config file if it is shallower.
  // The display path will be unaffected as this is based on workDirPath.
  if configFilePath != "" {
    configDirPath = filepath.Dir(configFilePath)
  }
}
...
```

This will let us drop the `validateOptions` helper, too.

internal/file/proto_set_provider.go
```diff
@@ -151,9 +162,15 @@ func (c *protoSetProvider) GetMultipleForFiles(workDirPath string, filePaths ...
...
 	filePathToProtoSet := make(map[string]*ProtoSet)
 	for dirPath, protoFiles := range dirPathToProtoFiles {
-		configFilePath, err := c.configProvider.GetFilePathForDir(dirPath)
-		if err != nil {
-			return nil, err
+		var configFilePath string
+		var err error
+		// we only want one ProtoSet if we have set configDirPath and configData
+		// since we are overriding all configuration files
+		if c.configDirPath == "" {
```
amckinney: We can update this to `c.configData == ""`

internal/file/proto_set_provider.go
```diff
@@ -164,8 +181,14 @@ func (c *protoSetProvider) getBaseProtoSets(dirPathToProtoFiles map[string][]*Pr
 		}
 		protoSet.DirPathToFiles[dirPath] = append(protoSet.DirPathToFiles[dirPath], protoFiles...)
 		var config settings.Config
-		// configFilePath is empty if no config file is found
-		if configFilePath != "" {
+		// we have already validated that configData is also set
+		if c.configDirPath != "" {
```
amckinney: We can update this to `c.configData != ""`

internal/file/proto_set_provider.go
```diff
@@ -195,6 +218,17 @@ func (c *protoSetProvider) walkAndGetAllProtoFiles(absWorkDirPath string, absDir
 		timedOut       bool
 	)
 	allExcludes := make(map[string]struct{})
+	// if we have a configDirPath and configData, we compute the exclude prefixes once
+	// from this dirPath and data, and do not do it again in the below walk function
+	if c.configDirPath != "" {
```
amckinney: We can update this to `c.configData != ""`

internal/file/proto_set_provider.go
```diff
@@ -212,12 +246,15 @@ func (c *protoSetProvider) walkAndGetAllProtoFiles(absWorkDirPath string, absDir
...
 				if fileInfo.IsDir() {
 					// Add the excluded files with respect to the current file path.
-					excludes, err := c.configProvider.GetExcludePrefixesForDir(filePath)
-					if err != nil {
-						return err
-					}
-					for _, exclude := range excludes {
-						allExcludes[exclude] = struct{}{}
+					// Do not add if we have configDirPath and configData.
+					if c.configDirPath == "" {
```
amckinney: We can update this to `c.configData == ""`

internal/file/proto_set_provider.go
```diff
@@ -266,6 +303,14 @@ func (c *protoSetProvider) walkAndGetAllProtoFiles(absWorkDirPath string, absDir
 	}
 }
 
+func (c *protoSetProvider) validateOptions() error {
+	if (c.configDirPath == "") != (c.configData == "") {
+		// this is a system error, this should not be possible for a user to configure
+		return fmt.Errorf("both configDirPath and configData in protoSetProvider must be set or unset")
+	}
+	return nil
+}
```
amckinney: This shouldn't be necessary (commented above).

internal/settings/config_provider.go
```diff
@@ -75,6 +75,15 @@ func (c *configProvider) Get(filePath string) (Config, error) {
...
 }
 
+func (c *configProvider) GetForData(dirPath string, externalConfigData string) (Config, error) {
+	var externalConfig ExternalConfig
+	if err := jsonUnmarshalStrict([]byte(externalConfigData), &externalConfig); err != nil {
+		return Config{}, err
+	}
+	dirPath = filepath.Clean(dirPath)
+	return externalConfigToConfig(externalConfig, dirPath)
+}
```
amckinney: Neat :+1:

internal/settings/config_provider.go
```diff
@@ -83,6 +92,18 @@ func (c *configProvider) GetExcludePrefixesForDir(dirPath string) ([]string, err
 	return getExcludePrefixesForDir(dirPath)
 }
 
+func (c *configProvider) GetExcludePrefixesForData(dirPath string, externalConfigData string) ([]string, error) {
+	if !filepath.IsAbs(dirPath) {
```
amckinney: Should we have this validation in `GetForData`, too?

internal/cmd/templates.go
peter-edge: This isn't global, by the same argument, `--protoc-url` should be global and we chose that not to be global.

peter-edge:
@AlexAPB perhaps we should sync offline - I don't get what you're concretely suggesting. The same amount of information is leaked from what I see, and changing the logic is a significant refactor. Maybe I'm missing something, but we should sync offline and/or I would want a concrete suggestion on how to proceed.

internal/cmd/flags.go
peter-edge: Hmmm, I disagree, let's cover this today offline?

internal/file/proto_set_provider.go
peter-edge: This logic is slightly wrong, you never use `absDirPath` as a value for `configDirPath` directly and the `else` is therefore not needed, but corrected it and updated.

peter-edge: Also note that this now requires propagating the work dir path to one more place, but we were already propagating it.

internal/file/proto_set_provider.go
peter-edge: Done

internal/file/proto_set_provider.go
peter-edge: Done

internal/file/proto_set_provider.go
peter-edge: Done

internal/file/proto_set_provider.go
peter-edge: Done

internal/settings/config_provider.go
peter-edge: Sure

PR #225, opened 2018-09-07
Add GitHub issue template

peter-edge:
I added a new GitHub Issue label "does not meet issue guidelines" to mark issues as such.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -15,12 +15,10 @@ Please do not @ individual maintainers in the description! We appreciate it.
 
 If your issue doesn't meet the guidelines, we'll comment as such and close the issue - we don't
 mean to be impersonal, but we have to reduce the support burden and cannot be supporting all
-of Protobuf (in the case of Protobuf questions or Protoc issues) or have longstanding issues
-without specifics (in the case of feature requests or bug reports). If the issue is a feature
-request and we don't think it meets the guidelines, don't feel like us closing the issue means
-we can't ever address it, unless we specifically note that we won't be adding a given feature
-(in which case the issue should stay closed) - provide more specifics and re-open the issue.
-If the issue is a bug report, unless we respond that Prototool is working as intended, add
-steps to reproduce the issue and re-open the issue.
+of Protobuf. If the issue is a feature request and we don't think it meets the guidelines,
+don't feel like us closing the issue means we can't ever address it, unless we specifically
+note that we won't be adding a given feature (in which case the issue should stay closed)
+- provide more specifics and re-open the issue. If the issue is a bug report, unless we respond
+that Prototool is working as intended, add steps to reproduce the issue and re-open the issue.
 
 Again, thanks for your input, and we hope Prototool helps your development!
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+  or ask your question on the Protobuf Mailing List at https://groups.google.com/forum/#!forum/protobuf.
+- If this is an issue with Protoc, do not file an issue here. File your issue
+  on the Protobuf repository at https://github.com/protocolbuffers/protobuf.
+- If this is a feature request, please provide a rough outline of the actual changes
+  you'd like to see with as many specifics as you can provide.
+- If this is a bug report, please provide specific steps to reproduce your issue.
+
+Please do not @ individual maintainers in the description! We appreciate it.
+
+If your issue doesn't meet the guidelines, we'll comment as such and close the issue - we don't
```
amckinney: A bit of a run-on. We can begin/end a new sentence here:
```
If your issue doesn't meet the guidelines, we'll comment as such and close the issue. We don't ...
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+mean to be impersonal, but we have to reduce the support burden and cannot be supporting all
+of Protobuf (in the case of Protobuf questions or Protoc issues) or have longstanding issues
```
amckinney: I would simply end this sentence at "but we have to reduce the support burden and cannot be supporting all of Protobuf." 

Then continue again at: "If the issue is a feature request ..."

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+without specifics (in the case of feature requests or bug reports). If the issue is a feature
```
amckinney: Can rephrase "If the issue is a feature ..." to "Unless we specifically note that we won't be moving forward with your feature request, a closed feature request means that it lacked sufficient context. Please provide more specifics and re-open the issue."

Thoughts?

.github/ISSUE_TEMPLATE.md
peter-edge: This one I actually sort of want to keep - the dash feels less aggressive, the new sentence feels aggressive. As an example:

This one I actually sort of want to keep. The dash feels less aggressive, the new sentence feels aggressive.

.github/ISSUE_TEMPLATE.md
peter-edge: Done

.github/ISSUE_TEMPLATE.md
peter-edge: Hmm this goes back to the aggressive/non-aggressive stance - I've purposefully made this message a bit more friendly, I don't want to give people a bad impression and also personally don't like it when people have issue templates that are similar to "Do this, or we will tell you to go away. Period. Full stop."

Does that makes sense?

peter-edge:
@amckinney I took one of the suggestions - I think we should get this in ASAP, but if you want to make more changes...well to quote the PR, "don't feel like us closing the issue means we can't ever address it" :-p Or better put, if you put up a small PR with any wording changes you want today, I'll aim to approve it within five minutes, regardless of the time zone difference :-)

PR #231, opened 2018-09-14
Add protoc-bin-path and protoc-wkt-path flags

amckinney:
This proposes two new flags: `--protoc-bin-path` and  `protoc-wkt-path`. This will let users effectively disable Prototool's caching feature. In general, caching is nice-to-have and desirable. However, Prototool needs to handle cases when the `protoc` binary is already compiled from another local file path (as opposed to being copied into its cache). This is especially relevant in build systems, such as [Buck](https://buckbuild.com/) or [Bazel](https://bazel.build/). 

Note that we purposefully add a separate `protoc-wkt-path` flag so that we guarantee that users are including the well-known types. An error is returned if either of these flags are set without the other.

I've tested this manually for the time being. If we think that this is the best approach forward, then I'm happy to add tests. I'd prefer to gather feedback before doing so because this is a potentially controversial change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Accept `prototool.json` files for configuation in addition to
   `prototool.yaml` files.
 - Add `--config-data` flag.
-- Add a `no-cache` flag to disable protoc caching.
+- Add `protoc-bin-path` and `protoc-wkt-path` flags to manually
+  set the paths for where `protoc` is run and where the
+  well-known types are included from.
 
 
 ## [1.2.0] - 2018-08-29
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Accept `prototool.json` files for configuation in addition to
   `prototool.yaml` files.
 - Add `--config-data` flag.
+- Add a `no-cache` flag to disable protoc caching.
 
 
 ## [1.2.0] - 2018-08-29
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -48,7 +48,7 @@ type flags struct {
 	pkg            string
 	printFields    string
 	protocBinPath  string
-	protocWktPath  string
+	protocWKTPath  string
 	protocURL      string
 	stdin          bool
 	uncomment      bool
@@ -146,8 +146,8 @@ func (f *flags) bindProtocBinPath(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.protocBinPath, "protoc-bin-path", "", "The path to the protoc binary. Setting this option will ignore the config protoc.version setting. This flag must be used with protoc-wkt-path.")
 }
 
-func (f *flags) bindProtocWktPath(flagSet *pflag.FlagSet) {
-	flagSet.StringVar(&f.protocWktPath, "protoc-wkt-path", "", "The path to the well-known types. Setting this option will ignore the config protoc.version setting. This flag must be used with protoc-bin-path.")
+func (f *flags) bindProtocWKTPath(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.protocWKTPath, "protoc-wkt-path", "", "The path to the well-known types. Setting this option will ignore the config protoc.version setting. This flag must be used with protoc-bin-path.")
 }
 
 func (f *flags) bindStdin(flagSet *pflag.FlagSet) {
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -44,6 +44,7 @@ type flags struct {
 	listLinters    bool
 	lintMode       bool
 	method         string
+	noCache        bool
 	overwrite      bool
 	pkg            string
 	printFields    string
@@ -122,6 +123,10 @@ func (f *flags) bindListLinters(flagSet *pflag.FlagSet) {
 
 func (f *flags) bindMethod(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.method, "method", "", "The GRPC method to call in the form package.Service/Method. This is required.")
+}
+
+func (f *flags) bindNoCache(flagSet *pflag.FlagSet) {
+	flagSet.BoolVar(&f.noCache, "no-cache", false, "Disable the Prototool cache for protoc.")
 }
 
 func (f *flags) bindOverwrite(flagSet *pflag.FlagSet) {
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -44,10 +44,11 @@ type flags struct {
 	listLinters    bool
 	lintMode       bool
 	method         string
-	noCache        bool
 	overwrite      bool
 	pkg            string
 	printFields    string
+	protocBinPath  string
+	protocWktPath  string
 	protocURL      string
 	stdin          bool
 	uncomment      bool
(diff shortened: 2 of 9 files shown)


peter-edge:
Probably easiest to chat offline, do you and @AlexAPB have time this morning?

amckinney:
@peter-edge Sounds good, we'll find a time that works for everyone. I've simplified the implementation a bit - we can discuss more in person.

internal/protoc/downloader.go
```diff
@@ -42,18 +42,27 @@ import (
 	"go.uber.org/zap"
 )
 
+const _filePrefix = "file://"
```
peter-edge: The `_` isn't used anywhere else in Prototool (or Golang, for that matter :-) ).

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Accept `prototool.json` files for configuation in addition to
   `prototool.yaml` files.
+- Add a `no-cache` flag to disable protoc caching.
```
amckinney: Will update Changelog here.

internal/cmd/flags.go
```diff
@@ -46,6 +46,8 @@ type flags struct {
 	overwrite      bool
 	pkg            string
 	printFields    string
+	protocBinPath  string
+	protocWktPath  string
```
peter-edge: WKT (I think this is what is done elsewhere right?)

internal/cmd/flags.go
```diff
@@ -135,6 +137,14 @@ func (f *flags) bindProtocURL(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&f.protocURL, "protoc-url", "", "The url to use to download the protoc zip file, otherwise uses GitHub Releases. Setting this option will ignore the config protoc.version setting.")
 }
 
+func (f *flags) bindProtocBinPath(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.protocBinPath, "protoc-bin-path", "", "The path to the protoc binary. Setting this option will ignore the config protoc.version setting. This flag must be used with protoc-wkt-path.")
```
peter-edge: Add that this flag and the other one cannot be used with the —protoc-url flag

internal/protoc/compiler.go
```diff
@@ -268,7 +270,10 @@ func (c *compiler) getCmdMetas(protoSet *file.ProtoSet) (cmdMetas []*cmdMeta, re
 	}()
 	// you need a new downloader for every ProtoSet as each configuration file could
 	// have a different protoc.version value
-	downloader := c.newDownloader(protoSet.Config)
+	downloader, err := c.newDownloader(protoSet.Config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create downloader: %v", err)
```
peter-edge: nit: you could just return the error, Prototool does not do aggressive error wrapping

internal/protoc/downloader.go
```diff
@@ -64,7 +71,25 @@ func newDownloader(config settings.Config, options ...DownloaderOption) *downloa
...
 	}
-	return downloader
+	if downloader.protocBinPath != "" || downloader.protocWktPath != "" {
+		if downloader.protocURL != "" {
+			return nil, fmt.Errorf("cannot use protoc-url in combination with either protoc-bin-path or protoc-wkt-path")
+		}
+		if downloader.protocBinPath == "" || downloader.protocWktPath == "" {
+			return nil, fmt.Errorf("both protoc-bin-path and protoc-wkt-path must be set")
+		}
+		cleanBinPath := filepath.Clean(downloader.protocBinPath)
```
peter-edge: Might want to check that this is executable, even better would be to make sure the version flag works as we check below

internal/protoc/downloader.go
```diff
@@ -64,7 +71,25 @@ func newDownloader(config settings.Config, options ...DownloaderOption) *downloa
...
+		if _, err := os.Stat(cleanBinPath); os.IsNotExist(err) {
+			return nil, err
+		}
+		cleanWktPath := filepath.Clean(downloader.protocWktPath)
```
peter-edge: Want to check that this is a directory and contains the directories google, google/protobuf

internal/protoc/downloader.go
```diff
@@ -64,7 +71,25 @@ func newDownloader(config settings.Config, options ...DownloaderOption) *downloa
...
+		if _, err := os.Stat(cleanWktPath); os.IsNotExist(err) {
+			return nil, err
+		}
+		downloader.protocBinPath = cleanBinPath
+		downloader.protocWktPath = cleanWktPath
+	}
+	return downloader, nil
```
peter-edge: Prototool probably has other things that return errors from constructors, but it is generally better practice to do all these checks within the functions, as is done now for other parts of the Downloader, and not updating the constructor to return error

internal/cmd/flags.go
amckinney: Yup, good call.

internal/cmd/flags.go
amckinney: :+1:

internal/protoc/downloader.go
```diff
@@ -71,23 +71,26 @@ func newDownloader(config settings.Config, options ...DownloaderOption) (*downlo
...
 		if _, err := os.Stat(cleanBinPath); os.IsNotExist(err) {
 			return nil, err
 		}
-		cleanWktPath := filepath.Clean(downloader.protocWktPath)
-		if _, err := os.Stat(cleanWktPath); os.IsNotExist(err) {
+		cleanWKTPath := filepath.Clean(downloader.protocWKTPath)
+		if _, err := os.Stat(cleanWKTPath); os.IsNotExist(err) {
+			return nil, err
+		}
+		if _, err := os.Stat(filepath.Join(cleanWKTPath, "google", "protobuf")); os.IsNotExist(err) {
```
peter-edge: It’s really easy to just also call fileInfo.IsDir() on the return value from stat

internal/protoc/downloader.go
amckinney: We can leverage the current output and don't need to add this into the logic here:
```
fork/exec /path/to/bin: permission denied
```

internal/protoc/downloader.go
amckinney: In this case, I'd prefer this to be determined up-front. This way, any repeated calls to `ProtocPath` and/or `WellKnownTypesIncludePath` don't need to re-validate. What do you think?

internal/protoc/downloader.go
amckinney: Sure, I can add that in!

internal/protoc/downloader.go
peter-edge: I don’t really care that much either way, this is fine if it works

internal/protoc/downloader.go
peter-edge: I guess if you want, I don’t think this is great error message and it’s a really easy check against the os.FileInfo, aka it’s what we effectively do in the Downloader if you don’t use this flag, so nice to be consistent

internal/protoc/compiler.go
```diff
@@ -498,10 +515,12 @@ func getIncludes(downloader Downloader, config settings.Config, dirPath string,
 		if err != nil {
 			return nil, err
 		}
-		includes = append(includes, wellKnownTypesIncludePath)
-		// TODO: not exactly platform independent
-		if strings.HasPrefix(dirPath, wellKnownTypesIncludePath) {
-			fileInIncludePath = true
+		if wellKnownTypesIncludePath != "" {
```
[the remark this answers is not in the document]
amckinney: Good catch 👍

amckinney:
I think I prefer to keep the `protoc` prefix since it's more clear that the well-known types are packaged alongside the `protoc` binary, i.e. the `.zip` ties them together. 

This is a similar stance to why we wanted to error if either flag was set without the other.

peter-edge:
Sounds fine to me

PR #242, opened 2018-11-15
Add support for unix domain sockets with --address flag

### LINKED ISSUE #241: Support unix sockets
## command I ran
```
prototool grpc myservice.proto \
  --address unix:///tmp/myservice.sock \
  --method myservice.FooService/DoFoo\
  --data '{"some": "data"}'
```

## what I expected to see
a successful response

## what I saw instead
```
dial tcp: address tcp////tmp/myservice.sock: unknown port
```

## discussion
the `--address` flag assumes a TCP address, but it'd be nice if a user can supply a `unix:///` prefix to allow for gRPC over unix socket

APTy:
Fixes #241

### DIFF
--- a/internal/grpc/handler.go
+++ b/internal/grpc/handler.go
@@ -72,7 +72,8 @@ func (h *handler) Invoke(fileDescriptorSets []*descriptor.FileDescriptorSet, add
 	if err != nil {
 		return err
 	}
-	clientConn, err := h.dial(address)
+	network, address := getNetworkAddress(address)
+	clientConn, err := h.dial(network, address)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/grpc/handler.go
```diff
@@ -94,10 +95,21 @@ func (h *handler) Invoke(fileDescriptorSets []*descriptor.FileDescriptorSet, add
 	return invocationEventHandler.Err()
 }
 
-func (h *handler) dial(address string) (*grpc.ClientConn, error) {
+func getNetworkAddress(address string) (string, string) {
```
peter-edge: Can we do this instead:

```go
func getNetworkAddress(address string) (string, string, error) {
  split := strings.SplitN(address, "://", 2)
  if len(split) != 2 {
    return "", "", fmt.Errorf("invalid address: %s", address)
  }
  switch split[0] {
    case "tcp", "unix":
      return split[0], split[1]
    default:
      return "", "", fmt.Errorf("invalid network, only tcp or unix allowed: %s", split[0])
  }
}
```

internal/grpc/handler.go
```diff
@@ -94,10 +95,21 @@ func (h *handler) Invoke(fileDescriptorSets []*descriptor.FileDescriptorSet, add
...
+	for _, network := range networks {
+		scheme := network + "://"
+		if strings.HasPrefix(address, scheme) {
+			return network, strings.TrimPrefix(address, scheme)
+		}
+	}
+	return "tcp", address
+}
+
+func (h *handler) dial(network, address string) (*grpc.ClientConn, error) {
```
peter-edge: Can you revert this to the original function signature and instead have this call `getNetworkAddress` instead?

internal/grpc/handler.go
APTy: +1

PR #247, opened 2018-12-17
Add linters for versioned packages

peter-edge:
This adds linters for packages that have a major version as their last value separated by `"."`s.

The linters:

- Verify that a package is of the form `package.vMAJORVERSION`
- Verify that a `go_package` value is of the form `packagevMAJORVERSION`.

This does not update the default linters. Follow-up PRs re-adding lint groups, and more formally specifying a "v2" style guide will be sent. Note nothing is finalized until we do a release off of dev (which we won't do for a while).


internal/lint/check_package_major_versioned.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	v.pkg = pkg
+}
+
+func (v *packageMajorVersionedVisitor) Finally() error {
+	if v.pkg == nil {
+		v.AddFailuref(scanner.Position{Filename: v.filename}, "No package declaration found.")
+		return nil
+	}
+	_, ok := protostrs.MajorVersion(v.pkg.Name)
```
peter-edge: nit: I violated my own Golang rule here, will update to scope within `if` block

PR #248, opened 2018-12-18
Add backward-compatibility checker

amckinney:
This introduces a new `compatible` package that can be used to compare two `*descriptor.FileDescriptorSets` against one another. The package considers a variety of possible changes that are not backward-compatible, such as removing a message's field.

The `compatible` package also includes the severity of the change (`source` and `wire`), such that it will distinguish between source code compatibility from wire compatibility. For reference, the checks are inspired by what is found in the [Proto3 language guide](https://developers.google.com/protocol-buffers/docs/proto3#updating). 

This also adds a `location` package that can be used to map the `*descriptor.SourceCodeInfo` to each of the relevant descriptor types. With this, we can recover what line and column caused each of the backward-incompatible changes.

The expected output is of the following form:
```
foo.proto:8:15:wire:Field "key" (1) was removed.
bar.proto:9:16:source:Field "value" (2) had its type updated from "int32" to "int64".
```

Note that this has been ported from an internal repository. Some of the unit tests are temporarily commented out and will be updated to conform to Prototool's package structure.

### DIFF
--- a/internal/compatible/check.go
+++ b/internal/compatible/check.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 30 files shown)


AlexAPB:
🔥

PR #250, opened 2018-12-20
Switch to using Golang modules support

### LINKED ISSUE #192: Move from glide to vgo and move internal/cmd/prototool back to cmd/prototool

peter-edge:
Two switches in a day. This switches dependencies to be tracked using [Golang modules](https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more). This is much faster, as well as allows an extremely simple install using `go get github.com/uber/prototool/cmd/prototool@VERSION`.

Fixes #192.

This removes yarpc and grpc-gateway from the example, as this was the only place these were used, as well as removes any internal generation using gogo/protobuf, instead using golang/protobuf which is what this library depends on internally.

This temporarily removes linting using errcheck, staticcheck, unused until they are resolved in https://github.com/golang/go/issues/24661, but the tradeoff to be able to use Golang modules is worth it IMO.


peter-edge:
Specific tracking issue for errcheck to work with Golang modules: https://github.com/kisielk/errcheck/issues/155

peter-edge:
Specific tracking issue for staticcheck, unused to work with Golang modules: https://github.com/dominikh/go-tools/issues/328

Issue #202, opened 2018-08-23
Install command to explicitly trigger a protoc install


lwc:
Hello, thanks for a great tool!

It's super handy being able to pin the version of `protoc` so easily, one thing we've noticed though is that `prototool` handles the install lazily/transparently. 

We're attempting to distribute a docker image with `protoc` and all of our required plugins baked in as a cohesive piece, and in order to get `prototool` to install `protoc`, we're doing this:

```dockerfile
ADD prototool.yaml /
# Required to make prototool install protoc for us
RUN prototool create temp.proto && prototool generate && rm temp.proto
```

It would be nice if there was a way to explicitly trigger `prototool` to install `protoc` in this case, e.g:
```dockerfile
ADD prototool.yaml /
RUN prototool install
```

Thanks!

peter-edge:
We used to have this and were waiting to see if a user would request it after we removed it heh. @alexapb @amckinney thoughts?

peter-edge:
@lwc I know this is super hacky, but this would accomplish what you need, including as a `RUN` directive for Docker:

```bash
rm -rf /tmp/prototool-bootstrap && \
  prototool config init /tmp/prototool-bootstrap && \
  echo 'syntax = "proto3";' > /tmp/prototool-bootstrap/tmp.proto && \
  prototool compile /tmp/prototool-bootstrap && \
  rm -rf /tmp/prototool-bootstrap
```

The better way to do this as part of a bash script would be to use `mktemp -d`, but for Darwin or Linux, the above should work. We could update the documentation to suggest this @amckinney @AlexAPB.

peter-edge:
Also see https://github.com/uber/prototool/pull/226

peter-edge:
Note if you want to re-add `prototool download`, you will want a way to specify the configuration to use. In the future, this might be stuff other than protoc, so a flag `--protoc-version` probably isn't the best. In this case, note that:

```bash
mkdir /tmp/prototool-bootstrap && \
  echo $'protoc:\n  version:' $PROTOC_VERSION > /tmp/prototool-bootstrap/prototool.yaml && \
  echo 'syntax = "proto3";' > /tmp/prototool-bootstrap/tmp.proto && \
  prototool compile /tmp/prototool-bootstrap
```

Will probably only be simplified to:

```bash
mkdir /tmp/prototool-bootstrap && \
  echo $'protoc:\n  version:' $PROTOC_VERSION > /tmp/prototool-bootstrap/prototool.yaml && \
[...]

PR #257, opened 2018-12-24
Cache commands

peter-edge:
Fixes https://github.com/uber/prototool/issues/202.

See the updates to the README.md and documentation for details.


README.md
```diff
@@ -462,35 +462,23 @@ The entire implementation is purposefully under the `internal` package to not ex
...
+# Basic invocation which will cache using the default behavior. See prototool help cache update for more details.
+prototool cache update
+# Cache to a specific directory path/to/cache
+prototool cache update --cache-path path/to/cache
+# Cache using custom configuration data instead of finding a prototool.yaml file using the file discovery mechanism
+prototool cache update --config-data '{"protoc":{"version":"3.6.1"}}'
 ```
 
-But for Darwin or Linux, the above should work.
+There is also a command `prototool cache delete` which will delete all cached assets of `prototool`,
```
konradjniemiec: I feel as if this behavior could lead to some frustrating behavior, unless they read this exact line of the README.

If you run:
```
prototool cache update --cache-path /tmp
prototool cache delete
prototool cache update --cache-path /tmp
```
You would expect the third command to download `protoc` again, however that won't be the case, the cache will be completely populated.

I understand the reasoning for keeping the option available, so perhaps the solution here is to add to the description of `--cache-path` that the user is expected to manage this folder, and to `cache delete` that it will only clean up the default cache.

internal/cmd/flags.go
```diff
@@ -62,7 +62,7 @@ func (f *flags) bindAddress(flagSet *pflag.FlagSet) {
 }
 
 func (f *flags) bindCachePath(flagSet *pflag.FlagSet) {
-	flagSet.StringVar(&f.cachePath, "cache-path", "", "The path to use for the cache, otherwise uses the default behavior.")
+	flagSet.StringVar(&f.cachePath, "cache-path", "", "The path to use for the cache, otherwise uses the default behavior. See prototool help cache update for more details.")
```
konradjniemiec:
```suggestion
	flagSet.StringVar(&f.cachePath, "cache-path", "", "The path to use for the cache, otherwise uses the default behavior. The user is expected to clean and manage this cache path. See prototool help cache update for more details.")
```

internal/cmd/templates.go
```diff
@@ -56,12 +57,39 @@ var (
...
-		Short: "Delete the cache.",
-		Args:  cobra.NoArgs,
+	cacheUpdateCmdTemplate = &cmdTemplate{
+		Use:   "update",
+		Short: "Update the cache by downloading all artifacts.",
+		Long: `This will download artifacts to a cache directory before running any commands. Note that calling this command is not necessary, all artifacts are automatically downloaded when required by other commands. This just provides a mechanism to pre-cache artifacts during your build.
+
+Artifacts are downloaded to the following directories based on flags and environment variables:
+
+- If --cache-path is set, then this directory will be used.
```
konradjniemiec:
```suggestion
- If a custom --cache-path is set, then this directory will be used, and is expected to be cleaned and managed by the user, i.e. prototool cache clean will not delete a custom cache path.
```

internal/cmd/templates.go
```diff
@@ -56,12 +57,39 @@ var (
...
+		},
+		BindFlags: func(flagSet *pflag.FlagSet, flags *flags) {
+			flags.bindCachePath(flagSet)
+			flags.bindConfigData(flagSet)
+		},
+	}
+
+	cacheDeleteCmdTemplate = &cmdTemplate{
+		Use:   "delete",
+		Short: "Delete all artifacts in the cache.",
```
konradjniemiec:
```suggestion
		Short: "Delete all artifacts in the default cache.",
```

internal/cmd/templates.go
```diff
@@ -56,12 +57,39 @@ var (
...
+		Long: `The following directory will be deleted based on environment variables:
+
+- If $XDG_CACHE_HOME is set, then $XDG_CACHE_HOME/prototool will be deleted.
+- Otherwise, if on Linux, $HOME/.cache/prototool will be deleted, or on Darwin,
+  $HOME/Library/Caches/prototool will be deleted.`,
```
konradjniemiec:
```suggestion
  $HOME/Library/Caches/prototool will be deleted.
  
  This will not delete any custom caches created using the --cache-path option.`,
```

README.md
peter-edge: I could see the confusion if users expected prototool to maintain some kind of config that stored flag invocations, but it doesn't and I wouldn't expect the second command to do anything to my custom cache path since I don't specify the path on `delete`. However, I added a line to the documentation about this.

internal/cmd/cmd_test.go
```diff
@@ -1008,7 +1008,7 @@ func testDownload(t *testing.T) {
 	// download checks if protoc is already downloaded to the cache location
```
smaye81: Should this test name change?  i.e. `testCacheUpdate`

peter-edge: This isn't actually a test, this just populates the cache, it's a helper function

fgrosse:
Great to see this just landet in `dev`. Any plans to release this soonish (e.g. as `v1.4.0`)?
I would otherwise probably switch to using the dev branch instead.

peter-edge:
See #264. Note dev is unstable

PR #258, opened 2018-12-26
Inspect commands

peter-edge:
This adds a bunch of inspection commands to print out data about packages. We can extend this in the future to include other items (messages within packages, files, etc).  We need most of this functionality anyways for the breaking change detector and there was an ask for it, so we might as well expose it.

Tldr:

```
$ cd example

$ prototool inspect packages
foo
google.protobuf
sub

$ prototool inspect packages --json
{
  "name": "foo",
  "deps": [
    "google.protobuf",
    "sub"
  ]
}
{
  "name": "google.protobuf",
  "importers": [
    "foo"
  ]
}
{
  "name": "sub",
  "importers": [
    "foo"
  ]
}

$ prototool inspect package --name foo
{
  "name": "foo",
  "deps": [
    "google.protobuf",
    "sub"
  ]
}

$ prototool inspect package-deps --name foo
google.protobuf
sub

$ prototool inspect package-importers --name sub
foo
```

### DIFF
--- a/internal/extract/extract.go
+++ b/internal/extract/extract.go
@@ -169,6 +169,7 @@ func (s sortPackages) Len() int          { return len(s) }
 func (s sortPackages) Swap(i int, j int) { s[i], s[j] = s[j], s[i] }
 func (s sortPackages) Less(i int, j int) bool {
 	if s[i] == nil && s[j] == nil {
+		return false
 	}
 	if s[i] == nil && s[j] != nil {
 		return true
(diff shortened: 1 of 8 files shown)


internal/extract/extract.go
```diff
@@ -88,3 +162,30 @@ func GetterWithLogger(logger *zap.Logger) GetterOption {
...
 	return newGetter(options...)
 }
+
+type sortPackages []*Package
+
+func (s sortPackages) Len() int          { return len(s) }
+func (s sortPackages) Swap(i int, j int) { s[i], s[j] = s[j], s[i] }
+func (s sortPackages) Less(i int, j int) bool {
+	if s[i] == nil && s[j] == nil {
+	}
```
smaye81: Should you return here?

PR #267, opened 2019-01-04
Finish MVP of reflect and extract packages

peter-edge:
This should be the majority of the functionality we need to support the initial breaking change detector. I'll add more testing as well later, but the actual function signatures may be in flux so I wanted to leave this as-is until the breaking change detector is done.


internal/reflect/proto/uber/proto/reflect/v1/reflect.proto
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// MessageOneof describes a Protobuf message oneof.
+message MessageOneof {
+  // name is the name of the message oneof.
+  string name = 1;
+  // field_numbers are the field numbers in the encapsulating type that make up
+  // this message oneof.
+  //
+  // This will ber sorted.
```
konradjniemiec:
```suggestion
  // This will be sorted.
```

PR #275, opened 2019-01-08
Move reflect testdata to reflect package

peter-edge:
Simple cleanup


peter-edge:
Cancelled build, Travis is buggy or something right now, it passed on PR which is what we care about

Issue #265, opened 2019-01-03
Linters to add for V2 Style Guide


peter-edge:
Running list to keep track:

- [ ] Outlaw public imports
- [ ] No reserved ranges or names, use deprecated instead

PR #281, opened 2019-01-10
Add linter to check for keywords in the package name

peter-edge:
Re: #265.


scooter-dangle:
@peter-edge, had you considered adding a larger catalog of keywords to the lint? We recently had an issue where some `FieldOptions` that were used to generate supplemental deserialization code for a Go application had a field added called `default` without realizing that it would break the code generated for a downstream Java application managed by a different team.

We're planning to add checks in CI against the keyword lists for Java, C, C++, Go, and JavaScript. We also use Rust and Python, but Rust's recent [addition of 'raw identifiers'](https://github.com/rust-lang/rust/pull/48942) lets it bypass the issue, and Python has [a workaround using `getattr`](https://stackoverflow.com/a/38061021/808850) (although I don't know whether that's considered ergonomic).

Of course, the trade-off is the linter having to enumerate many languages' keywords, which might be an unjustifiable maintenance burden.

PR #291, opened 2019-01-11
Fail break check if a stable package depends on a beta package unless --allow-beta-deps is set

peter-edge:
Re: #265

### DIFF
--- a/internal/breaking/breaking.go
+++ b/internal/breaking/breaking.go
@@ -118,6 +118,14 @@ var (
 			Check:   checkServiceMethodsSameServerStreaming,
 		},
 	}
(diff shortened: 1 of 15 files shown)


internal/breaking/runner.go
```diff
@@ -55,22 +56,24 @@ func (r *runner) Run(from *extract.PackageSet, to *extract.PackageSet) ([]*text.
 			return nil, err
 		}
 	}
+	checkers := r.checkers
+	// if includeBeta, do not do the check
```
konradjniemiec: nit: make the comment a sentence.

PR #298, opened 2019-01-13
Add linters to verify that enums, messages, services, and service methods have complete sentence comments

peter-edge:
These linters do not require Golang style `// TypeName ...`, but do require that the comments start with an uppercase letter and contain a period.


peter-edge:
Re: #265.

PR #337, opened 2019-02-13
Glob matching for lint ignore files

pbarker:
Adds support for glob matching lint ignore files. The prototool config file becomes overly verbose when needing to explicitly list all files.

- Adds dep on https://github.com/bmatcuk/doublestar 
In our use case, pattern matching is most useful when you can recursively exclude directories. Go has a long standing issue for support https://github.com/golang/go/issues/11862, but it appears to be dragging.

### DIFF
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -7,6 +7,11 @@ def prototool_deps(**kwargs):
         importpath = "honnef.co/go/tools",
     )
     go_repository(
+        name = "com_github_bmatcuk_doublestar",
+        importpath = "github.com/bmatcuk/doublestar",
+        tag = "v1.1.1",
+    )
+    go_repository(
         name = "com_github_burntsushi_toml",
         importpath = "github.com/BurntSushi/toml",
         tag = "v0.3.1",
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/uber/prototool
 
 require (
+	github.com/bmatcuk/doublestar v1.1.1
 	github.com/cpuguy83/go-md2man v1.0.8 // indirect
 	github.com/emicklei/proto v1.6.7
 	github.com/fullstorydev/grpcurl v1.1.0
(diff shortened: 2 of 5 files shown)


peter-edge:
Why do you need to list files? If you have a linter that needs ignoring globally, add it to `lint.rules.remove`.

There are some issues with this PR:

- bmatcuk/doublestar is not widely adopted, and looking at the code briefly, there are some issues, for example the code would not pass `golint`. We have some libraries that aren't widely adopted as dependencies, but we want to avoid adding more.
- `lint.ignores.files` is a non-obvious place to have wildcards. You'd probably have to add a new `patterns` sub-option, and then handle that throughout the code.
- The documentation for Prototool is not updated.
- There is no testing.

I'm not comfortable with this change, and before the above could be fixed, I'd need to see why this is necessary - if you're doing a lot of ignores with `lint.ignores.files`, it sounds like you need to instead use `lint.rules.remove` as you aren't widely adopting the rule set.

peter-edge:
Please feel free to comment here and I'll reply, but as I don't see this being merged, I'm going to close this for now. Sorry!

pbarker:
Hey @peter-edge, let me try and enumerate the use case a bit. For various reasons related to the ecosystem, we need to exclude whole paths from certain linters. In doing so our exclude rules have blown up, as we're doing this for many files and many linters.

The simple solve here is to allow globing, which seems generally expected when dealing with filepaths. In our case, because one proto lives per directory having double star functionality makes sense, but as you see in the go issue they still haven't supported that internally. However, even just supporting the internal `path/filepath` glob functionality would cut down on a fair chunk of the bloat.

I'm happy to add tests and docs if this looks like something that could be merged. Although, I didn't see tests for `lint.go`, but I could be misunderstanding how you're doing that. 

Thanks for the consideration!

peter-edge:
Why don't you just set the `excludes` option?

pbarker:
So, I still need to generate from the protos, they just don't adhere to the others in regards to linting. I could separate them in to different root directories and use different config files, but it wouldn't make sense from a repo structure perspective, and I would then have a fair amount of duplication in the config files.

peter-edge:
Ah got it. OK, fair enough - here's what I need:


- Documentation needs to be updated. 
- I'd prefer if we could just document this to be the same as `excludes`, i.e. you can either specify a directory, in which case everything under this directory will be ignored, or you can specify a single file. This is what I need to update `excludes` documentation to reflect. Actually, while I'm complaining about this, I'm being hypocritical, I need to update the documentation for `excludes` to better reflect what it is:

Would you be OK with just being able to specify directories to exclude or files to exclude, with no wildcards? What this would look like:

```
lint:
  ignores:
    - id: FOO
      files:
        - path/to/file.proto
      dirs:
        - path/to/dir
```

This is a bit of an involved change, do you mind if I just do it? It will only take me like 15 mins but the PR could go back and forth a bit.

pbarker:
yep that looks awesome, and don't mind who does it. Thanks for the help!

peter-edge:
Filed an issue for it, will get to it soon

peter-edge:
See https://github.com/uber/prototool/pull/346

peter-edge:
Fixed on dev

PR #349, opened 2019-02-25
Added 'regexp' plugin option

andviro:
Hi there! First of all, thanks for immensely useful tool, really simplifies protobuf workflow. The only thing I missed is an ability to selectively apply different plugins to different directories with .proto's. E.g. I'd like to generate swagger files for a small portion of services, not all of them. So here we are: new option for a plugin, "regexp". If specified, only protosets whose paths are matched are processed by the plugin. As an added bonus, it is possible to have different records in plugins' section with the same name that match different portions of source tree.

Best regards,
A.

### DIFF
--- a/etc/config/example/prototool.yaml
+++ b/etc/config/example/prototool.yaml
@@ -134,6 +134,12 @@ generate:
       # The Mfile=package flags are automatically set.
       # ** Otherwise, generally do not set this unless you know what you are doing. **
       flags: plugins=grpc
+      # Regular expression pattern.
+      # If specified, .proto absolute path should match the pattern,
+      # otherwise plugin is not applied.
+      # Multiple plugins with the same name and different match patterns
+      # and configurations can be specified.
+      regexp: .
 
       # The path to output generated files to.
       # If the directory does not exist, it will be created when running generation.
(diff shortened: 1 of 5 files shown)


peter-edge:
I don't know if I can be convinced on this one - the idea with Prototool is that you have a consistent generated set of stubs regardless of language, and the end-user does not have to think about what is generated where. If we did do this, it would need testing.

Why cant you just do a post-processing step to delete the files you don't want after `prototool` runs?

peter-edge:
We don't allow regular expressions for anything else - `excludes, includes`, lint ignores - I don't think we should special case plugin generation here. I could perhaps be convinced on an `exclude` string list that excludes specific directories for plugins, but even that makes me slightly uncomfortable - perhaps though, I'd need to think about it a bit.

andviro:
Fair enough. But to have specific plugins/plugin flags for different parts of protoset would be very useful. I tried the least intrusive approach, albeit a little hack-y. What I'd actually like more is ability to specify 'directory', similar to 'generate' section. That would be really cool.

peter-edge:
If you have a concrete proposal there, file an issue and we can discuss the layout there before you implement it, is that OK? I don't think we're going to merge `regexp` specifically for plugins.

andviro:
Sure. Will do, thanks for your time.

PR #360, opened 2019-03-03

yinzara:
As the "prototool grpc" command is just a passthrough to the grpcurl library, there was no way to pass the attributes to the tool for TLS related connections.  By default the current use of the library would be with the '-plaintext' flag (as it does not pass a TLS creds object to "BlockingDial" call).  

This PR adds new command line flags to the "grpc" command to support TLS connections.  These flags mirror the same flags from the grpcurl cli.  The only difference relates to the "--tls" flag itself (which doesn't exist on the grpcurl library).  Since "prototool grpc" by default implies the "-plaintext" flag from grpcurl, a new "--tls" flag will exist that enables TLS (the reverse of --plaintext).

### DIFF
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -26,6 +26,12 @@ import (
 
 type flags struct {
 	address        string
+	tls            bool
+	insecure       bool
+	cacert         string
+	cert           string
+	key            string
+	serverName     string
 	cachePath      string
 	callTimeout    string
 	configData     string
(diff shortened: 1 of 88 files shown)


internal/cmd/flags.go
```diff
@@ -161,3 +167,28 @@ func (f *flags) bindUncomment(flagSet *pflag.FlagSet) {
...
+
+func (f *flags) bindCert(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.cert, "cert", "", "File containing client certificate (public key), to present to the server. Ignored if --tls is not specified. Must also provide the --key.")
+}
+
+func (f *flags) bindKey(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.key, "key", "", "File containing client private key, to present to the server. Ignored if --tls is not specified. Must also provide the --cert option.")
+}
+
+func (f *flags) bindSeverName(flagSet *pflag.FlagSet) {
```
peter-edge: Typo?

yinzara: YES! Totally my mistake, fixing and resubmitting.

yinzara:
Looks like the build wants me to also check in all files created with "make generate" which will update nearly every file in the repo with a new copyright date.

peter-edge:
This should not be against master, this should be against dev.

yinzara:
I'm happy to cancel this PR and move it to the "dev" branch if you'd like. 

Regarding flag names, other than "--tls", the flag names match one for one to the 'grpcurl' command's flags itself.  If we want to add a layer of abstraction and rename them, we can do that, but I feel like the original names are pretty clear.

peter-edge:
It's adding five flags, we need to look over it. We use grpcurl as a library, we don't match it's CLI flags.

peter-edge:
#216 (not merged) only used one flag - we probably don't need to expose everything here.

yinzara:
I respectfully disagree.  

Just because you don't currently have a use case that uses the other parameters, this tool is meant to be generic for use for others, the other parameters exist for a reason and the pull request should take all into account.  

My use case wouldn't even be covered by the other pull request (i.e. I do not need to provide a certificate authority file, but I need to enable TLS because my service utilizes certificates issued by LetsEncrypt that are accepted by system approved CAs) so in my case I only use "--tls" parameter however in his case he would use "--tls" and "--cacert" parameter.  However originally my team, contemplated requiring client certificates themselves (which would require the --cert and --key parameters) while others may not want the complexity of dealing with cert validation (especially in a test environment) and would want to pass "--tls" and "--insecure".

My solution utilizes the exact same code used by the "grpcurl" to build the TransportCredentials object, the other pull request uses some kinda of custom method for building the credentials.

peter-edge:
We add features as they are needed - you can always add more, but since we're 1.0, we can't remove features once they are added.

If you want to put up a PR against dev, we can evaluate it.

yinzara:
What determines "as they are needed" and "by whom"? 

The "prototool grpc" command is meant to support sending gRPC requests to a gRPC service. Since gRPC services allow TLS connections and many even require it (some even requiring a client certificate), it is already "needed" in my opinion, since there are use cases that require it.

The other person who submitted the pull request "needed" the CA certificate functionality. I "needed" it to support TLS without supplying a CA.  Other may "need" it to allow for insecure TLS connections if they can't guarantee the CA.

PR #362, opened 2019-03-05

yinzara:
As the "prototool grpc" command is just a passthrough to the grpcurl library, there was no way to pass the attributes to the tool for TLS related connections. By default the current use of the library would be with the '-plaintext' flag (as it does not pass a TLS creds object to "BlockingDial" call).

This PR adds new command line flags to the "grpc" command to support TLS connections. These flags mirror the same flags from the grpcurl cli. The only difference relates to the "--tls" flag itself (which doesn't exist on the grpcurl library). Since "prototool grpc" by default implies the "-plaintext" flag from grpcurl, a new "--tls" flag will exist that enables TLS (the reverse of --plaintext).

Originally referenced as PR #360 , this PR merges to "dev" instead of "master"

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,31 @@ $ cat input.json | prototool grpc example \
 }
 ```
 
+### SSL/TLS Connections
+
+To enable SSL/TLS connections to the server, use the `--tls` command line flag. 
+If not specified all the following flags will be ignored.
+
+By default host validation is enabled. To disable, pass the `--insecure` command line flag.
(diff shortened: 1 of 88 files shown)


peter-edge:
You still have this based off the master branch, you just did the PR against dev - this is why there are conflicting files.

yinzara:
I'm going to open this under a feature branch in my fork instead of master after rebasing.

PR #364, opened 2019-03-05

yinzara:
As the "prototool grpc" command is just a passthrough to the grpcurl library, there was no way to pass the attributes to the tool for TLS related connections. By default the current use of the library would be with the '-plaintext' flag (as it does not pass a TLS creds object to "BlockingDial" call).

This PR adds new command line flags to the "grpc" command to support TLS connections. These flags mirror the same flags from the grpcurl cli. The only difference relates to the "--tls" flag itself (which doesn't exist on the grpcurl library). Since "prototool grpc" by default implies the "-plaintext" flag from grpcurl, a new "--tls" flag will exist that enables TLS (the reverse of --plaintext).

--tls  - enable SSL/TLS all together.  If just this is specified, the system supported certificate authorities will be used and a SSL/TLS connection will be attempted.  This is specifically the use case I need (my certificates are generated from LetsEncrypt).  This is also the flag to enable all the other flags below. Technically we could infer --tls if any of the below are specified however I liked the idea of being able to turn off tls completely by omitting this flag and keeping the other flags (maybe we could generate a warning or something).  I also thought about using "--ssl" but technically that's not what it is anymore (we're all TLS now).  This is the easiest option to support secure gRPC connections as long as your server uses a certificate issued by a globally trusted CA.  This is the default when using "grpcurl" from the command line.

--cert and --key are the SSL/TLS certificate and private key which are used when client transport encryption is required
See https://bbengfort.github.io/programmer/2017/03/03/secure-grpc.html
They are respectively server.crt (or server.pem) and server.key from this site.

--cacert (--certificate-file|--ssl-certificate-path|--certificate from PR #216 ) the naming from 216 is confusing as technically this is not the SSL certificate but a file containing the public certificate of the trusted certificate authorities. This would be "rootca.crt" from the above tutorial however if your cert is published from a system trusted certificate authority, you don't need to specify this.

--insecure   For those that can't have a truly validated SSL/TLS connection (i.e. they don't have all of the above items but still need SSL/TLS), this flag disables host validation but still uses TLS/SSL.  This is useful in a testing environment but should never be used in production.

I'm certainly happy to discuss changing the flag names (as I've based them 100% on grpcurl and I'm not sure you want to tie us to that library completely) however they are all required if we would like to support the full range of gRPC connection options.

I'm using the grpcurl.ClientTransportCredentials function to create the credentials object which requires the 4 flags I added as part of this PR however we could manually create the TransportCredential in the same manner the grpcurl library does.

Originally submitted as PR #360 but moved to dev branch.


README.md
```diff
@@ -177,6 +177,31 @@ Call a gRPC endpoint using a JSON input. What this does behind the scenes:
 
 *See [docs/grpc.md](docs/grpc.md) for full instructions.*
 
+### SSL/TLS Connections
```
peter-edge: Can you put this in `docs/grpc.md` instead?

README.md
```diff
@@ -177,6 +177,31 @@ Call a gRPC endpoint using a JSON input. What this does behind the scenes:
...
+
+To enable SSL/TLS connections to the server, use the `--tls` command line flag. 
+If not specified all the following flags will be ignored.
+
+By default host validation is enabled. To disable, pass the `--insecure` command line flag.
```
peter-edge: `--tls --insecure` doesn't make sense to me - having a `--insecure` flag to use as a TLS option seems counterintuitive. Perhaps `--skip-host-validation` or similar, open to suggestions.

README.md
```diff
@@ -177,6 +177,31 @@ Call a gRPC endpoint using a JSON input. What this does behind the scenes:
...
+
+To use a custom certificate authority, pass the `--cacert` command line flag to the path of the file in pem format.
+
+To use a custom key and certificate, pass the `--key` and `--cert` command line flag to the path of the files in pem format.
+
+To use a custom server name for validation, pass `--server-name` command line flag.
```
peter-edge: What happens if I don't use this flag?

All flags need their default behavior explained. https://godoc.org/github.com/fullstorydev/grpcurl#ClientTransportCredentials has some explanation on the defaults, but we need a full overview here.

README.md
```diff
@@ -177,6 +177,31 @@ Call a gRPC endpoint using a JSON input. What this does behind the scenes:
 
 *See [docs/grpc.md](docs/grpc.md) for full instructions.*
 
+### SSL/TLS Connections
+
+To enable SSL/TLS connections to the server, use the `--tls` command line flag.
```
peter-edge: There's no equivalent mechanism in any other command, and this doesn't actually enable or disable the other flags as you say in the PR description - they are all installed regardless. I'm not sure if this is needed.

If we need this after discussion, then instead of ignoring fields below, they should error if set. So for example `--server-name` should result in an error if `--tls` is not set.

internal/cmd/flags.go
```diff
@@ -201,3 +207,28 @@ func (f *flags) bindUncomment(flagSet *pflag.FlagSet) {
 func (f *flags) bindFix(flagSet *pflag.FlagSet) {
 	flagSet.BoolVarP(&f.fix, "fix", "f", false, "Fix the file according to the Style Guide.")
 }
+
+func (f *flags) bindTLS(flagSet *pflag.FlagSet) {
+	flagSet.BoolVar(&f.tls, "tls", false, "Enable SSL/TLS connection to remote host")
```
peter-edge: See comments, but if we keep this, this needs a period

internal/exec/runner.go
```diff
@@ -511,6 +511,9 @@ func (r *runner) GRPC(args, headers []string, address, method, data, callTimeout
 	if data != "" && stdin {
 		return newExitErrorf(255, "must set only one of data or stdin")
 	}
+	if (cert != "") != (key != "") {
+		return newExitErrorf( 255, "If cert is specified, key must be specified")
```
peter-edge: do not capitalize errors

internal/grpc/handler.go
```diff
@@ -113,6 +124,23 @@ func getNetworkAddress(address string) (string, string, error) {
 	}
 }
 
+func (h *handler) getClientTransportCredentials() (credentials.TransportCredentials, error) {
+	var creds credentials.TransportCredentials
```
peter-edge: This should be:

```
if !h.tls {
  return nil, nl
}
```

To match the rest of this repo, this no need to declare creds as a var

internal/cmd/flags.go
```diff
@@ -201,3 +207,28 @@ func (f *flags) bindUncomment(flagSet *pflag.FlagSet) {
...
+func (f *flags) bindInsecure(flagSet *pflag.FlagSet) {
+	flagSet.BoolVar(&f.insecure, "insecure", false, "Disable host certificate validation for TLS connections. Ignored if --tls is not specified.")
+}
+
+func (f *flags) bindCacert(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.cacert, "cacert", "", "File containing trusted root certificates for verifying the server. Ignored if --tls is not specified.")
+}
+
+func (f *flags) bindCert(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.cert, "cert", "", "File containing client certificate (public key), to present to the server. Ignored if --tls is not specified. Must also provide the --key.")
```
peter-edge: Documentation should match the documentation for `--key` in structure

internal/exec/runner.go
```diff
@@ -498,7 +498,7 @@ func (r *runner) All(args []string, disableFormat, disableLint, fixFlag bool) er
 	return nil
 }
 
-func (r *runner) GRPC(args, headers []string, address, method, data, callTimeout, connectTimeout, keepaliveTime string, stdin bool) error {
+func (r *runner) GRPC(args, headers []string, address, method, data, callTimeout, connectTimeout, keepaliveTime string, stdin bool, tls bool, insecure bool, cacert string, cert string, key string, serverName string) error {
```
peter-edge: See discussion on `--tls` flag. We would need to verify that flags are all correctly set, not just ignore them.

peter-edge:
Can you also run `make` locally? We've temporarily disabled Bazel testing on CI because it takes too long - we'll be fixing this before v1.4.0, but right now we want to verify that `make` passes locally.

README.md
yinzara: I actually completely agree.  It was succinct.  Chef uses "--ssl-verify-mode none".  Wget uses "--no-check-certificate".  Curl (which is why the grpcurl library uses):
-k, --insecure

(TLS) By default, every SSL connection curl makes is verified to be secure. This option allows curl to proceed and operate even for server connections otherwise considered insecure.

The server connection is verified by making sure the server's certificate contains the right name and verifies successfully using the cert store.

See this online resource for further details: https://curl.haxx.se/docs/sslcerts.html

README.md
yinzara: Absolutely, once we agree on final flags I'll move it there.

README.md
yinzara: I'll update the readme when I move it to the suggested location and incorporate this comment.

README.md
yinzara: We need the option to exist because of the simple case (i.e. TLS connection with default CA's trusted) so I like the option of an error if --tls is not set yet one of the other 4 parameters are specified.

internal/cmd/flags.go
yinzara: noted.

internal/exec/runner.go
yinzara: k noted

internal/grpc/handler.go
yinzara: got it. ty

internal/exec/runner.go
yinzara: Ok I agree with your prior comment. I think producing an error when the tls is not set yet other parameters are specified.

README.md
peter-edge: How does grpcurl handle this? It doesn't have a `--tls` flag equivalent.

yinzara:
Yes, I did run make locally and it did pass. 

Initially it did not.  I believe the resolution was the installation of the following packages into my GOPATH:
go get github.com/uber-go/zap
go get github.com/stretchr/testify
go get github.com/jhump/protoreflect
go get github.com/emicklei/proto
go get github.com/fullstorydev/grpcurl
go get github.com/uber-go/multierr 

I think I also needed to install Bazel locally (I received cryptic errors without it installed)

README.md
yinzara: It assumes it by default. You have to pass a -plaintext flag to it for it to behave the way prototool grpc does now.  We could make prototool grpc function that way but it would change the default behavior of the tool.

peter-edge:
Are you using the Makefile commands, or running commands outside of the Makefile? If you are using the Makefile, the above shouldn't be necessary - the Makefile automatically sets `GO111MODULE=on` and downloads it's own copy of Bazel.

README.md
peter-edge: Got it. OK, let's go with the `--tls` flag and then error if TLS-specific flags are set without the TLS flag set (and document that they fail without the TLS flag)

yinzara: The primary idiosyncrasy between `grpcurl` and `curl` is that curl requires the transport protocol (i.e. https, http etc) where grpcurl does not.  Because of that, it has to be instructed to use a secure connection or not via some other flag.  By default grpcurl assumes a TLS connection and needs the -plaintext flag to function with an unencrypted connection.

yinzara: I think there is kind of a fundamental decision here. Do you want "prototool grpc" to function analogously with "curl" and therefore you should probably name all the flags the same (as grpcurl does) or do you want to create your own abstraction and only rely on grpcurl as the current implementation.

peter-edge: Note that right now, you can actually set an address as either `tcp://address` or `unix://address` (not well documented) - we could expand this to accept `http` or `https` and use that as the mechanism, but honestly I think I prefer the `--tls` flag having thought about this a bit more.

peter-edge: `prototool grpc` isn't functionally equivalent to either, it was just a last-minute addition before we released `prototool` - I'm actually working to document how to use other grpc-specific CLI tools with prototool for cases where grpc reflection isn't available (but this is a big TODO, and I want to add the ability to generate "mega-FileDescriptorSets" first).

That prototool uses grpcurl is just an implementation detail, and if we wanted, we might rip out grpcurl in the future if something else came along.

internal/grpc/handler.go
yinzara: So if I do that I still need to declare the var:
`
        if !h.tls {
		return nil, nil
	}
	var err error
	var creds credentials.TransportCredentials
	creds, err = grpcurl.ClientTransportCredentials(h.insecure, h.cacert, h.cert, h.key)
	if err != nil {
		return nil, err
	}
	if h.serverName != "" {
		if err := creds.OverrideServerName(h.serverName); err != nil {
			return nil, err
		}
	}
	return creds, nil
`

yinzara:
I used "make", but it was giving me errors like:
Open the coverage report:
open .tmp/Darwin/x86_64/etc/coverage.html
bazel build //...:all
ERROR: /Users/yinzara/go/src/github.com/uber/prototool/internal/format/BUILD.bazel:3:1: no such package '@org_uber_go_zap//': no such package '@bazel_gazelle_go_repository_tools//': failed to build fetch_repo: build cache is required, but could not be located: GOCACHE is not defined and $HOME is not defined
 and referenced by '//internal/format:go_default_library'
ERROR: Analysis of target '//internal/format:go_default_library' failed; build aborted: no such package '@org_uber_go_zap//': no such package '@bazel_gazelle_go_repository_tools//': failed to build fetch_repo: build cache is required, but could not be located: GOCACHE is not defined and $HOME is not defined
INFO: Elapsed time: 0.921s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (36 packages loaded, 32 targets configured)
    Fetching @com_github_stretchr_testify; Restarting.
    Fetching @com_github_gobuffalo_flect; Restarting.
    Fetching @org_uber_go_zap; Restarting.
    Fetching @org_golang_google_grpc; Cloning df014850f6dee74ba2fc94874043a9f3f75fbfd8 of https://github.com/grpc/grpc-go
    Fetching @com_github_golang_protobuf; Cloning c823c79ea1570fb5ff454033735a8e68575d1d0f of https://github.com/golang/protobuf
    Fetching @com_github_fullstorydev_grpcurl; Restarting.
    Fetching @com_github_jhump_protoreflect; Restarting.
    Fetching @com_github_gofrs_flock; Restarting. ... (16 fetches)


When I installed each library referenced in the error with the go get commands above, it worked fine.

peter-edge:
Ah shoot, this is actually a bug related to Golang 1.12. Will put up a fix.

README.md
yinzara: How about this for a validation check:
```
	if tls {
		if (cert != "") != (key != "") {
			return newExitErrorf(255, "if cert is specified, key must be specified")
		} else if insecure && (cacert != "" || cert != "" || key != "" || serverName != "") {
			return newExitErrorf(255, "if insecure then cacert, cert, key, and server-name must not be specified")
		}
	} else if insecure || cacert != "" || cert != "" || key != "" || serverName != "" {
		return newExitErrorf(255, "tls must be specified if insecure, cacert, cert, key or server-name are specified")
	}
```

yinzara: or should we include the "--" with each flag in the errors:
```
	if tls {
		if (cert != "") != (key != "") {
			return newExitErrorf(255, "if --cert is specified, --key must be specified")
		} else if insecure && (cacert != "" || cert != "" || key != "" || serverName != "") {
			return newExitErrorf(255, "if --insecure then --cacert, --cert, --key, and --server-name must not be specified")
		}
	} else if insecure || cacert != "" || cert != "" || key != "" || serverName != "" {
		return newExitErrorf(255, "--tls must be specified if --insecure, --cacert, --cert, --key or --server-name are specified")
	}
```

yinzara: The only other way to support this without a flag would be to infer it based upon the prefix of the address. Right now we infer "tcp://" if the address doesn't have "tcp://" or "unix://" at the beginning. We could also allow one or all of "https" or "tls" or "ssl" or "ssh" to imply "tcp" with "tls".

But then the question becomes, should we also infer it if any of the other 4 flags are passed if no prefix is given in the address. What happens if the prefix is "tcp"?

peter-edge: Yea let's just do the flag and not touch this part.

peter-edge:
If you merge from `dev`, the build problem should be fixed.

peter-edge:
I just merged #375 which causes conflicts here, you'll want to merge from dev again. Let me know when you have the second iteration of this PR up post-comments, I'll take a look!

yinzara:
If I had known how much work that was gonna be to get it done right, I don't know if I ever would have started ;-) It's 3am here and I've been working for 9 hours straight so forgive me.

Firstly I incorporated all of the changes you suggested.

I also improved the documentation in the grpc readme quite extensively to help people better understand how to use the flags.

I have now also added an entire suite of TLS related GRPC tests that I believe covers all of the functionality I added including server only TLS and mutual TLS (and all error scenarios as well).

The only scenario I can't actually test is one involving a CA that is in the system default CAs and verifying a valid certificate against that (as I would have to have a certificate generated by one of those CAs).  I do have a test for a failure to validate the system CA.

Because the tests required actually generating TLS certificates, I added a README in  internal/cmd/testdata/grpc that describes how they were generated (and how to regenerate them if you ever wanted to... but you shouldn't need to as they are valid for 100 years).

peter-edge:
This is some extremely impressive work, I really appreciate it.

There's a couple things I want to do to finish this up so you don't have to work anymore, namely make the generation of the certs automatic in the Makefile (`certstrap` is a Golang project so easy to install, and `openssl` I can add a check for), and to also change the name of the `--insecure` flag. Do you mind giving me write access on your fork and I'll push the changes when I get a chance?

Again, super appreciate the work here.

yinzara:
To be perfectly honest, it's the first ever Golang project I've worked on or submitted a PR for.  Little bit of a learning experience but all in all, pretty fun.

You have been granted write access.

peter-edge:
Great, I’ll get on it soon. Thanks again.

peter-edge:
@yinzara I made minimal modifactions, mostly did some cleanups. Again, this is great. Thanks so much for the effort. I'm going to have one more team member look over it and then we will merge.

docs/grpc.md
```diff
@@ -68,3 +70,46 @@ $ prototool grpc example \
...
+certificate against the given CA certificate (generally only true in commercially generated certificates,
+not self signed ones).
+
+By default, the "Common Name" (and any "Subject Alternative Names") will be used to validate the authenticity of the
+server connection based upon the `--address` specified.  If your server certificate's common name does not match the
+address you are accessing it from (for example, if your address is an IP address), pass `--server-name` command line
+flag with the value expected in the "Common Name" field of the server certificate.
+
+To use mutual TLS with a client key and certificate, pass the `--key` and `--cert` command line flag to the path of
+the files in pem format. If one is specified the other must also be specified. Unlike curl, you cannot pass a file with
```
konradjniemiec: nit: extra newlines on this line and next

internal/cmd/flags.go
```diff
@@ -225,3 +231,27 @@ func (f *flags) bindUncomment(flagSet *pflag.FlagSet) {
...
+func (f *flags) bindCacert(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.cacert, "cacert", "", "File containing trusted root certificates for verifying the server. Can also be a file containing the public certificate of the server itself. If set, --tls is required.")
+}
+
+func (f *flags) bindCert(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.cert, "cert", "", "File containing client certificate (public key) in pem encoded format to present to the server for mutual TLS authentication. If set, --tls and --key is required.")
+}
+
+func (f *flags) bindKey(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&f.key, "key", "", "File containing client key (private key) in pem encoded format to present to the server for mutual TLS authentication. If set, --tls and --cert is required.")
```
konradjniemiec:
```suggestion
	flagSet.StringVar(&f.key, "key", "", "File containing client key (private key) in pem encoded format to use for mutual TLS authentication. If set, --tls and --cert is required.")
```
nit: the private key isn't `presented to the server` it is used to sign messages for the mutual TLS authentication.

PR #390, opened 2019-03-11
Add non-default RPCS_NO_STREAMING linter

peter-edge:
This isn't included in any lint group by default, but will be available to add if someone wants it.


peter-edge:
Another failure based on https://github.com/golang/go/issues/30690. Merging anyways.

PR #399, opened 2019-03-16
gofmt -s + fix typo

chemidy:
* gofmt -s
* fix typo "specified"


peter-edge:
This would need to be integrated as part of the build before we would merge. CI also needs to pass. If you’d like to do the work to make this part of the build, I can consider this

However, I don’t think we need to do -s, I think the explicitness of the types is fine.

peter-edge:
I'll merge this for now, I'll potentially integrate `gofmt` into the build separately. Thanks!

PR #401, opened 2019-03-17
Remove all references to proto.EnumField.ValueOption

### LINKED ISSUE #400: Format: enum values lose options > 1
When more than one option is provided to an enum value, `prototool format` will strip out everything after the first one.

Example proto:
```protobuf
enum SomeEnum {
  SOME_ENUM_UNKNOWN = 0 [
    (option_1) = "a_value",
    (option_2) = "another_value"
  ];
}
```

From a quick look, this appears to be using [`EnumField.ValueOption`](https://godoc.org/github.com/emicklei/proto#EnumField) rather than `EnumField.Elements` in the format logic.

peter-edge:
Fixes #400.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,7 @@ grpcgen: $(CERTSTRAP)
 
 .PHONY: generate
 generate: license golden example internalgen bazelgen
+	gofmt -s -w $(SRCS)
 	go mod tidy -v
 
 .PHONY: checknodiffgenerated
@@ -174,7 +175,6 @@ errcheck: $(ERRCHECK)
 .PHONY: staticcheck
 staticcheck: $(STATICCHECK)
 	staticcheck --tests=false $(PKGS)
-
 
 .PHONY: checklicense
 checklicense: $(UPDATE_LICENSE)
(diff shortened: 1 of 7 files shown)


cvanderschuere:
Thanks for the incredibly quick fix!

One comment:
It [looks](https://github.com/uber/prototool/search?utf8=%E2%9C%93&q=ValueOption&type=) like ValueOption is also used in a couple of the linters. They don't seem to be causing any trouble, but probably worth addressing as well.

I believe it should be something like this (caveat: very unfamiliar with this library)
```golang 
if element.ValueOption != nil {
  element.ValueOption.Accept(v)
}
```
should become:

```golang 
for _, elem := range element.Elements {
   elem.Accept(v)
}
```

peter-edge:
Good point - I should have audited this. I'll look around now.

The `Elements` struct is slightly weird with regards to enum value options, so I'll probably handle it in a similar manner in other places just to make sure existing code isn't touched as much as possible.

peter-edge:
I removed the remaining references to `ValueOption`. I also snuck in `gofmt -s -w` re: #399.

cvanderschuere:
LGTM

peter-edge:
Going to wait for one of the other maintainers to just double-check and then will merge. Sorry for the wait.

PR #420, opened 2019-04-03
Add --generate-ignores flag to lint

peter-edge:
From the documentation:

You can configure ignoring of lint rules on a per-file basis:

```yaml
lint:
  ignores:
    - id: MESSAGE_NAMES_CAMEL_CASE
      files:
        - foo.proto
        - bar/baz.proto
```

To generate the a YAML configuration for currently-failing lint rules that can be copied into your
configuration file, use `--generate-ignores`. This will lint your files, ignoring the existing
setting for `lint.ignores`, and print a new value for it. Note that you should make sure not to
touch other settings for `lint` in your configurartion file as this flag only generates the
`lint.ignores` option.

```
prototool lint path/to/dir --generate-ignores
```

### DIFF
--- a/docs/lint.md
+++ b/docs/lint.md
@@ -65,7 +65,7 @@ lint:
 To generate the a YAML configuration for currently-failing lint rules that can be copied into your
 configuration file, use `--generate-ignores`. This will lint your files, ignoring the existing
 setting for `lint.ignores`, and print a new value for it. Note that you should make sure not to
-touch other settings for `lint` in your configurartion file as this flag only generates the
+touch other settings for `lint` in your configuration file as this flag only generates the
 `lint.ignores` option.
 
 ```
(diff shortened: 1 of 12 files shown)


docs/lint.md
```diff
@@ -51,6 +51,27 @@ lint:
...
+    - id: MESSAGE_NAMES_CAMEL_CASE
+      files:
+        - foo.proto
+        - bar/baz.proto
+```
+
+To generate the a YAML configuration for currently-failing lint rules that can be copied into your
+configuration file, use `--generate-ignores`. This will lint your files, ignoring the existing
+setting for `lint.ignores`, and print a new value for it. Note that you should make sure not to
+touch other settings for `lint` in your configurartion file as this flag only generates the
```
smaye81: nit:  `configuration`

peter-edge: Done

PR #438, opened 2019-04-23
added protoc-gen-govalidators to the docker image

vladshub:
Added integration with [go-proto-validators](https://github.com/mwitkow/go-proto-validators) to the docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,8 @@ RUN GO111MODULE=on go get \
   github.com/gogo/protobuf/protoc-gen-gogo@v${GOGO_PROTOBUF_VERSION} \
   github.com/gogo/protobuf/protoc-gen-gogofast@v${GOGO_PROTOBUF_VERSION} \
   github.com/gogo/protobuf/protoc-gen-gogofaster@v${GOGO_PROTOBUF_VERSION} \
-  github.com/gogo/protobuf/protoc-gen-gogoslick@v${GOGO_PROTOBUF_VERSION} && \
+  github.com/gogo/protobuf/protoc-gen-gogoslick@v${GOGO_PROTOBUF_VERSION} \
+  github.com/mwitkow/go-proto-validators/protoc-gen-govalidators && \
   mv /go/bin/protoc-gen-go* /usr/local/bin/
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN GO111MODULE=on go get \
   github.com/gogo/protobuf/protoc-gen-gogo@v${GOGO_PROTOBUF_VERSION} \
   github.com/gogo/protobuf/protoc-gen-gogofast@v${GOGO_PROTOBUF_VERSION} \
   github.com/gogo/protobuf/protoc-gen-gogofaster@v${GOGO_PROTOBUF_VERSION} \
-  github.com/gogo/protobuf/protoc-gen-gogoslick@v${GOGO_PROTOBUF_VERSION} && \
+  github.com/gogo/protobuf/protoc-gen-gogoslick@v${GOGO_PROTOBUF_VERSION} \
+  github.com/mwitkow/go-proto-validators/protoc-gen-govalidators && \
```
peter-edge: This is not versioned

go.mod
```diff
@@ -8,10 +8,12 @@ require (
 	github.com/fullstorydev/grpcurl v1.2.1
 	github.com/gobuffalo/flect v0.1.1
 	github.com/gofrs/flock v0.7.1
+	github.com/gogo/protobuf v1.2.1 // indirect
 	github.com/golang/protobuf v1.3.1
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/jhump/protoreflect v1.2.0
 	github.com/mitchellh/go-wordwrap v1.0.0
+	github.com/mwitkow/go-proto-validators v0.0.0-20190212092829-1f388280e944 // indirect
```
peter-edge: This should not be here, this is not part of the dependencies for the binary, same with above

vladshub:
@peter-edge this was an extremely fast response. 
Thank you for the comment.
I'll have a docker image in docker hub that builds upon yours and adds this binary until they will release a proper version

peter-edge:
OK sounds good! That's the recommended approach, I'll update the documentation to reflect this. Thanks for your understanding.

PR #466, opened 2019-06-09
Update README.md

Inilien:
added suggestion to use company name as the first sub-name of a package.


smaye81:
Thank you for filing, but we don’t want to add this to our style guide.

PR #479, opened 2019-07-29
Ignore the vscode directory

smaye81:
Ignores the `.vscode` directory


bweston92:
You can add this to your own global ignore file.

smaye81:
> You can add this to your own global ignore file.

Sure, but that doesn't prevent anyone else from committing vscode artifacts into the repo.

PR #487, opened 2019-08-21
Update sorting of Failures

micnncim:


smaye81:
Thank you for the PR, but I'm not entirely sure if this is necessary.  I don't see a reason to make this change.

micnncim:
I also think so... I'll close this issue, thank you!

PR #500, opened 2019-09-25
Add flag for generators that otherwise overwrite.

### LINKED ISSUE #484: protoc-gen-doc files are overwritten if there are multiple proto files in the directory
I have a module with two proto files in the directory. I have the below configuration for the "doc" plugin. 
`
    - name: doc
      flags: markdown,test.md
      output: docs
`
when I invoke prototool, I see it protoc command twice (once for each proto file). This causes the second run to overwrite the doc generated from the first one. Is there a workaround for this?

mattbailey:
Adds NamespaceOutput (namespace_output: bool in YAML) to create
directories that map to proto file directories in order to overcome
generators like protoc-gen-doc that do not create subdirectories, and
end up overwriting their output file.

Fixes #484

Maintainers:
- `make` auto updated `bazel/deps.bzl`, let me know if that's not desired
- This PR is partly to ask if I'm on the right track in how you want to attack this problem, so I haven't added more documentation other than comments (and no examples). Will add them if you think this is a good solve.


smaye81:
Thank you for being patient on this.  I think this actually may be a bit more difficult than anticipated.  I think I'm going to close this for now and keep #484 open until we can address it a bit better.

PR #502, opened 2019-10-01
Add protoc-gen-swift to Docker image

surajbarkale:
This PR adds `protoc-gen-swift` tool from https://github.com/apple/swift-protobuf to the docker image. Since swift does not support static linking, I had to explicitly copy all .so dependencies into the final image. Since all other tools are statically linked, copying over glibc works.

I have tested locally by adding generate clause in `prototool.yaml`:
```yaml
generate:
  plugins:
    - name: swift
      output: swift
```
Please let me know if any explicit tests have to be added.

I understand this PR may be rejected because of future conflicts with glibc. However, this will help someone who is using prototool and needs to generate swift bindings.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,7 +68,7 @@ RUN cd /tmp/prototool && \
 RUN upx --lzma /usr/local/bin/*
 
 FROM swift:5.1 AS swift-builder
-ENV SWIFT_PROTO_VERSION=1.7.0
+ENV SWIFT_PROTO_VERSION=1.8.0
 RUN set -ex; \
   mkdir -p /tmp/swift-proto; \
   git clone -c advice.detachedHead=false --depth 1 -b ${SWIFT_PROTO_VERSION} \
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,8 +109,8 @@ ENV \
   PROTOTOOL_PROTOC_BIN_PATH=/usr/bin/protoc \
   PROTOTOOL_PROTOC_WKT_PATH=/usr/include \
   GRPC_VERSION=1.21.3 \
-  PROTOBUF_VERSION=3.8.0 \
-  ALPINE_GRPC_VERSION_SUFFIX=r0 \
+  PROTOBUF_VERSION=3.9.2 \
+  ALPINE_GRPC_VERSION_SUFFIX=r1 \
   ALPINE_PROTOBUF_VERSION_SUFFIX=r0
 
 RUN echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,22 @@
-FROM golang:1.12.4-alpine3.9 as builder
+ARG GOLANG_PROTOBUF_VERSION=1.3.1
+ARG GOGO_PROTOBUF_VERSION=1.2.1
+ARG GRPC_GATEWAY_VERSION=1.8.5
+ARG GRPC_WEB_VERSION=1.0.4
+ARG YARPC_VERSION=1.37.3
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -67,6 +67,40 @@ RUN cd /tmp/prototool && \
...
 RUN upx --lzma /usr/local/bin/*
 
+FROM swift:5.1 AS swift-builder
+ENV SWIFT_PROTO_VERSION=1.7.0
+RUN set -ex; \
+  mkdir -p /tmp/swift-proto; \
+  git clone -c advice.detachedHead=false --depth 1 -b ${SWIFT_PROTO_VERSION} \
+    https://github.com/apple/swift-protobuf.git /tmp/swift-proto; \
+  cd /tmp/swift-proto; \
+  swift build -c release; \
```
jan-xyz: If you compile the plugin with a static stdlib you should not need to copy the other files.

```suggestion
  swift build --static-swift-stdlib  -c release; \
```

orkunkl:
Hello! What is the current status of this PR? This feature would be really cool.

smaye81:
Apologies for the delay.  I have been a bit out of the loop/busy.  I will take a look at this tonight.  Thank you for putting this up.

Dockerfile
surajbarkale: It does remove a couple of libraries at the expense of binary size. I have added the switch nevertheless. I have also automated the file copying to make this less brittle

smaye81:
Looks good.  Can you run `make dockerall` to verify the container builds correctly and that the Docker tests all work?

surajbarkale:
I have added tests and fixed issues with `make dockerall`. However, there is now a substantial change in the Dockerfile.

surajbarkale:
I also found out that statically linked swift executables do not work on Linux!
https://bugs.swift.org/browse/SR-648

smaye81:
Yeah so I'd really rather not add this level of complexity to the Dockerfile all for the sake of adding in Swift.  Our preference is for you to build your own image based on `uber/prototool` anyway.

See here:  https://github.com/uber/prototool/blob/dev/docs/docker.md

surajbarkale:
Here is the gist of changes I have made to the image:

1. Bumped GRPC version to 1.25.0 and protobuf version to 3.11.2. Running `make dockerall` on master is failing because the existing versions are no longer available in alpine:edge.

2. Use alpine:3.11 instead of alpine:edge. Since alpine 3.10 contains older version of GRPC and protobuf than master.

3. Changed go version to 1.13 since the go.mod file in master specifies this version.

4. Move all version definitions to start of the Dockerfile and reuse these in build steps. This is a cleanup and makes it easier to check build versions.

5. Split builder image into base & builder image. Otherwise the swift builder was always re-run.

6. Added swift builder image as an independent image.

Changes 1, 2 & 3 are needed to fix `make dockerall` on master. Change 4 is an enhancement (you may disagree). Changes 5 & 6 bring in swift support.

I am documenting this here in case anyone wants to reuse code in this PR. Please feel free to close this PR if you don't want to merge it.

smaye81:
I think Steps 1-4 are valid.  I am in the process of actually upgrading dependencies now.  The `dev` branch should have most `go.mod` and `protoc` updates, but the Bazel build is failing, so I need to work that out.  So, I would say Steps 1-3 I can take care of.  

Step 4 is also a valid thing.  I'd definitely welcome a PR of that if possible.

And yeah, I think I'd rather not do Steps 5-6 right now.  Apologies, but at least you can use that code to build your own image based on the Prototool image.

smaye81:
Going to close this PR.  If you decide to implement Step 4 in our above convo, please open a new PR.

jnvillar:
Is this working? It created some swift files for me but I am not able to find the services

PR #519, opened 2019-11-01
Add config file path

syndbg:
Since prototool does its best to find configuration files inside included dirs (why does it do that in the first place?), specifying the config file path makes it possible to use prototool in projects that have the following hierarchy:

1) protobuf definitions git repo
2) SDK for X language that includes `1)` as git submodule
3) project using `2`.

P.S The contribution flow is not entirely clear. Is this supposed to be against `dev`? I've based my base on `dev`. What's your release flow?

### DIFF
--- a/internal/cmd/flags.go
+++ b/internal/cmd/flags.go
@@ -31,6 +31,7 @@ type flags struct {
 	cacert            string
 	cert              string
 	configData        string
+	configFilePath    string
 	connectTimeout    string
 	data              string
 	debug             bool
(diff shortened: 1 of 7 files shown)


sashayakovtseva:
+1 for this flag

However, testing on the following hierarchy didn't work:
```
repo/
  api/
    api.proto
  configs/
    prototool.yaml
```

Without changing `prototool.yaml` (which I had in repo root),  I tried to move it to the `configs` dir and ran the following command:  `prototool generate --config-file-path configs/prototool.yaml`. Nothing was generated :(

syndbg:
@sashayakovtseva Sure, I'll check out this case.

smaye81:
Is this still being investigated?

syndbg:
Hi, no. Forgot about it 🤡 I'll check it out tomorrow

jathu:
@syndbg friendly bump. This would be really useful :)

smaye81:
Closing due to inactivity.  I also think there is more to this effort than this PR addresses, but if anyone else wants to take this on, feel free to reopen a new PR.

PR #521, opened 2019-11-12
Don't discard error

GrimTheReaper:
This error is being shadowed. Maybe it should be checked?


smaye81:
Fixed on dev as part of https://github.com/uber/prototool/pull/540

PR #524, opened 2019-11-18
Add protoc-gen-mypy to Docker image

juzna:
The mypy compiler (https://github.com/dropbox/mypy-protobuf/) generates typing information, which is very useful for IDEs and static analysis.

Also upgrading to newest versions of dependencies in Alpine in separate commit, because pinned versions are not available anymore, so the image can't be built currently.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,13 +74,12 @@ WORKDIR /work
 ENV \
   PROTOTOOL_PROTOC_BIN_PATH=/usr/bin/protoc \
   PROTOTOOL_PROTOC_WKT_PATH=/usr/include \
-  GRPC_VERSION=1.21.3 \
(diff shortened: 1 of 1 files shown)


smaye81:
Thank for being patient on this.  If you are still interested in this PR, can you rebase dev and update any outdated deps in the Dockerfile?  If not, I can just close the PR.

smaye81:
Closing.  If this is still needed, let's discuss.  We'd prefer just building your own image off of `uber/prototool` anyway.

juzna:
Yes, we ended up using our own protool image. I just thought that mypy is so generic that it would be useful to many.

smaye81:
Nice, I'm glad you were able to work it out.  Thanks for putting up the original PR, though.  Glad you found a solution.

PR #529, opened 2019-12-11
Add support of ENV vars in the config-provider

demdxx:
To be able to use more flexible file paths was added support of ENV vars in the config-provider.

The main idea is the possibility to use the include paths from the system environment.

For example:

```yml
protoc:
  version: 3.8.0

  includes:
    - ".."
    - ${GOPATH}/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.12.1/third_party/googleapis
    - ${GOPATH}/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.12.1

lint:
  file_header:
    content: |
      // ${USER} - Document generation
    is_commented: true
```

Also, was extended `generate.go_options.import_path`, `lint.file_header.path`, `lint.file_header.content` and `lint.ignores.files` with the same reason.

NOTE: Originally it was implemented only for several fields of config but can be extended for the whole config file.

### DIFF
--- a/internal/settings/config_provider.go
+++ b/internal/settings/config_provider.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -204,6 +205,7 @@ func externalConfigToConfig(develMode bool, e ExternalConfig, dirPath string) (C
 	}
 	includePaths := make([]string, 0, len(e.Protoc.Includes))
 	for _, includePath := range strs.SortUniq(e.Protoc.Includes) {
+		includePath = processEnvVariables(includePath, os.Getenv)
 		if !filepath.IsAbs(includePath) {
 			includePath = filepath.Join(dirPath, includePath)
 		}
@@ -214,6 +216,7 @@ func externalConfigToConfig(develMode bool, e ExternalConfig, dirPath string) (C
 	for _, ignore := range e.Lint.Ignores {
 		id := strings.ToUpper(ignore.ID)
 		for _, protoFilePath := range ignore.Files {
+			protoFilePath = processEnvVariables(protoFilePath, os.Getenv)
 			if !filepath.IsAbs(protoFilePath) {
 				protoFilePath = filepath.Join(dirPath, protoFilePath)
 			}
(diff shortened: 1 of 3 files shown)


internal/settings/config_provider.go
```diff
@@ -433,6 +437,32 @@ func jsonUnmarshalStrict(data []byte, v interface{}) error {
...
+	return envVariableRegexp.ReplaceAllStringFunc(str, func(s string) string {
+		if strings.HasPrefix(s, "${") {
+			return mapper(s[2 : len(s)-1])
+		}
+		varIndex := 0
+		for i := 0; i < len(s); i++ {
+			if s[i] != '$' {
+				break
+			}
+			varIndex++
```
smaye81: This is a pretty strange way of doing the replacement.  Can we make this better?  Also, I'm not sure I agree with the replacement rules:

Why does `$${PWD}` become `${PWD}` but `$$${PWD}` gets the replacement string with an extra `$` on the end?  That doesn't make a lot of sense.

demdxx: Yes, it's not the best way of "char-escaping". Was chosen SQL escape approach like ('''') => ('') as for configs used YAML and JSON which already supports backslash escaping "\n\t\\" what means that to escape the same way need to write "\\$" which is also not really pretty, but it's OK to me to use the "\\$" if you don't mind.
Or if you have other suggestions please share it to me

smaye81: I really feel like this opens up a bit of complexity with resolving paths and variables, etc. that I'd rather not introduce to Prototool at this point.  I apologize.

I would recommend maybe forking the repo and using that fork with your env var changes there.

demdxx: To use it from my repository a bit harder the original. I think that using the global environment in paths is easier and convenient in many cases, especially it does not affect any other internal logic.

As you can see from the example it makes my life much easier. So if you will decide the way of using global environment variables I will glad to implement it for you and for my self.

```sh
PROTOTOOL_VERSION := efdb8d5776a679a0032cfb2c81e17270aa629941
PROTOTOOL := $(TMP_VERSIONS)/prototool/$(PROTOTOOL_VERSION)
$(PROTOTOOL):
	# $(eval PROTOTOOL_TMP := $(shell mktemp -d))
	# cd $(PROTOTOOL_TMP); go get github.com/uber/prototool@$(PROTOTOOL_VERSION)
	# @rm -rf $(PROTOTOOL_TMP)
	@rm -rf $(TMP)/src/prototool
	@mkdir -p $(TMP)/src/prototool
	cd $(TMP)/src/prototool; git clone https://github.com/demdxx/prototool .; \
		git checkout $(PROTOTOOL_VERSION); go build -o ./prototool cmd/prototool/main.go
	rm -f $(TMP_BIN)/prototool
	mv $(TMP)/src/prototool/prototool $(TMP_BIN)/prototool
	@rm -rf $(dir $(PROTOTOOL))
	@mkdir -p $(dir $(PROTOTOOL))
	@touch $(PROTOTOOL)
	@rm -rf $(TMP)/src/prototool
```

smaye81: Sorry for the delay, but again, I don't want to put anything regarding env vars in Prototool right now.  Why is it harder for you to fork and implement the changes there?

smaye81:
Closing due to inactivity and the fact this is something I'd rather not implement in Prototool

PR #534, opened 2020-01-10
Add linter for ruby_package file option

jathu:
This PR adds a linter to ensure `ruby_package` file option is set. This file option is [part of protobuf](https://github.com/protocolbuffers/protobuf/blob/5a578d2a5cd3bce73b4b4a4ebd9ba40ff7fa1931/src/google/protobuf/descriptor.proto#L429).


jathu:
@smaye81 done https://github.com/uber/prototool/pull/534/commits/dc896f2bad3c0e6ffd499cc9e3b2f1f5369f5ef3

smaye81:
You will also need to add it to one of the lint groups (`Uber2Linters`).  You only added it to the `AllLinters` slice.

jathu:
Wouldn't that apply this rule to the uber2 lint group? Do we want to enforce that?

smaye81:
Yeah I think that's fine.  We will just have to document that it's a new rule in the `uber2` lint group.

jathu:
@smaye81 cool, I've added the rule to the `uber2` lint group. It's grown the scope of the PR a bit, but I think I've covered all the affected tests.

smaye81:
Fyi, you will need to rebase the `dev` branch.  Tests should pass after the rebase.  I landed a few dependency updates.

smaye81:
Apologies, but after talking with someone, they made me realize that adding to `uber2` will break everyone's linting with a minor release.  So, I think I'd like to just add the linter but not add it to the `uber2` group.  Can you revert the changes that add it to the group?

jathu:
@smaye81 reverted 👍

